PR #132, opened 2026-04-12
Fix Codex P1: wire Android accessibility + Windows SDF include

danielraffel:
Addresses Codex P1 comments from PRs #129 and #131.


danielraffel:
Resolved by PR #151 (commit edb8dbbb): `pulp-jni` now force-references each `Java_com_pulp_accessibility_*` symbol via `-Wl,--undefined=` so the linker can't strip them. Verified Android build green on both macos-latest and windows-latest hosts. (PR #148 attempted whole-archive on pulp-view first but caused a duplicate JNI_OnLoad collision with SDL3.)

PR #141, opened 2026-04-12
MPE Phase 4: MIDI 2.0 UMP native path (closes #139)

### LINKED ISSUE #139: MPE Phase 4: MIDI 2.0 UMP native path
## Context

Phases 1-3 of MPE support shipped in #135 (MpeVoiceTracker, MpeBuffer sidecar with CLAP wiring, MpeSynthVoice/MpeVoiceAllocator/MpeGlideDetector, `examples/mpe-synth/`, docs, and `pulp create --mpe`).

Phase 4 — the MIDI 2.0 UMP native path — was intentionally deferred. This issue tracks what needs to land when we pick it up.

## Goals

MIDI 2.0 UMP carries per-note pitch bend, per-note CCs, and 16-bit velocity natively, without the MPE-over-MIDI-1.0 channel-steering trick. Today Pulp parses UMP packets (see `core/midi/include/pulp/midi/ump.hpp`) but does not route UMP-native data through the plugin interface — everything flows through `MidiBuffer` (MIDI 1.0 status bytes).

Phase 4 adds a first-class UMP path so hosts/transports that speak MIDI 2.0 (CLAP's future UMP events, CoreMIDI UMP, Android AMidi 2.0, WebMIDI 2.0) can deliver native packets to plugins without lossy down-conversion.

## Tasks

- [ ] **`UmpBuffer`** — parallel to `MidiBuffer`, holds a sample-ordered stream of `UmpPacket`s. Lock-free on the audio thread (preallocated capacity, no per-event allocation).
- [ ] **`MidiBuffer` ↔ `UmpBuffer` bidirectional conversion** — a `midi1_to_ump()` that emits equivalent MIDI 1.0 UMP packets (message type 0x2) for any MIDI 1.0 event, and a `ump_to_midi1()` that flattens MIDI 2.0 per-note messages into MIDI 1.0 with MPE-style channel steering when the processor doesn't support UMP.
- [ ] **Processor sidecar** — `Processor::ump_input()` / `set_ump_input(const UmpBuffer*)`, mirroring the existing `mpe_input()` pattern. Opt-in via a new `PluginDescriptor::supports_ump = true` flag (keeps `supports_mpe` for plugins that only want the MIDI 1.0 sidecar).
- [ ] **Format-adapter plumbing** — CLAP first (it will receive `CLAP_EVENT_MIDI2` once upstream ships it; until then, synthesize UMP by running inbound MIDI 1.0 through `midi1_to_ump()`). VST3 and AU adapters use the same fallback.
- [ ] **`MpeVoiceTracker` UMP ingestion** — accept `UmpPacket` input alongside `MidiEvent`, using MIDI 2.0 per-note pitch bend (`0x60`) and per-note CCs (`0x00`/`0x10`) directly instead of the member-channel heuristic.
- [ ] **Round-trip tests** — MIDI 1.0 → UMP → MIDI 1.0 equivalence for common channel-voice messages; UMP 16-bit velocity preserved through the sidecar; per-note pitch bend preserved.
- [ ] **Conversion accuracy tests** — 14-bit MIDI 1.0 pitch bend ↔ 32-bit UMP pitch bend scaling; 7-bit ↔ 16-bit velocity scaling per the MIDI 2.0 spec.
- [ ] **Example** — extend `examples/mpe-synth/` (or a new `examples/ump-synth/`) to opt into `supports_ump` and consume the native stream when present, falling back to the MPE sidecar otherwise.
- [ ] **Docs** — expand `docs/guides/mpe.md` (or a new `docs/guides/ump.md`) to document the dual `MpeBuffer` / `UmpBuffer` story and when to pick which.

## Out of scope

- **MIDI-CI profile negotiation upgrades** beyond what already exists in `core/midi/include/pulp/midi/midi_ci.hpp`. Those are a separate follow-up.
- **Sysex8 / data128 handling** — Phase 4 focuses on channel-voice and per-note messages; bulk-data transports can come later.
- **Physical UMP transport plumbing** (CoreMIDI UMP endpoints, AMidi 2.0) — wire these per-platform once the core types are in place.

## Design notes

- Keep `UmpPacket` as the source of truth; `UmpBuffer` just gives it a block-scoped container with sample offsets.
- Prefer one-way construction on the audio thread: adapters fill `UmpBuffer` before `process()`, the processor reads it, and the adapter clears it after. No in-place mutation contracts.
- Reuse the existing `MpeVoiceTracker` state machine where possible — the tracker already owns per-channel/per-note state, so UMP ingestion becomes a second input path rather than a second tracker.
- The `MpeBuffer` sidecar stays; plugins that want MPE-semantics can keep using it regardless of whether the transport is MIDI 1.0 or MIDI 2.0.

## References

- Source plan: `planning/next-features-plan.md` § Feature 2 Phase 4
- Ralph prompt: `planning/ralph-prompt-mpe.md` (notes Phase 4 as deferred)
- Existing UMP types: `core/midi/include/pulp/midi/ump.hpp`
- MPE sidecar precedent (Phase 2): `core/midi/include/pulp/midi/mpe_buffer.hpp`, `core/format/src/clap_adapter.cpp`

danielraffel:
## Summary

Implements Phase 4 of MPE support: a MIDI 2.0 UMP opt-in sidecar parallel to the existing MPE sidecar. Plugins that declare \`supports_ump = true\` receive a sample-accurate \`UmpBuffer\` via \`Processor::ump_input()\` during \`process()\`. \`MpeVoiceTracker\` now also ingests \`UmpPacket\`s, routing per-note pitch bend and per-note CCs directly to the matching note (full MIDI 2.0 resolution, no channel-steering heuristic).

The CLAP adapter populates the sidecar by converting inbound MIDI 1.0 events with \`midi1_to_ump()\` until CLAP ships \`CLAP_EVENT_MIDI2\`; plugins see a stable API either way.

Closes #139.

## Test plan
- [x] \`pulp-test-ump-buffer-conversion\` — 10 cases / 45 assertions, incl. scaling round-trips, centre preservation, MIDI 1.0 ↔ UMP equivalence, UMP per-note bend isolation
- [x] Existing MPE tests still green (tracker / buffer / synth voice / mpe-synth)
- [x] Full local build + CLAP dlopen for PulpMpeSynth
- [ ] Shipyard ship — mac + Namespace Linux/Windows

### DIFF
--- a/core/format/include/pulp/format/clap_adapter.hpp
+++ b/core/format/include/pulp/format/clap_adapter.hpp
@@ -44,6 +44,12 @@ struct PulpClapPlugin {
     int32_t mpe_current_sample_offset = 0;
     bool mpe_enabled = false;
 
+    // UMP sidecar — populated by converting midi_in to MIDI 2.0 UMP packets
+    // when the Processor declares supports_ump. In the future, CLAP hosts
+    // that deliver CLAP_EVENT_MIDI2 can populate this buffer directly.
+    midi::UmpBuffer ump_buffer;
+    bool ump_enabled = false;
+
     // Preset management (optional — set by plugins that provide presets)
     std::unique_ptr<state::PresetManager> preset_manager;
 
--- a/core/format/include/pulp/format/processor.hpp
+++ b/core/format/include/pulp/format/processor.hpp
@@ -3,6 +3,7 @@
 #include <pulp/audio/buffer.hpp>
 #include <pulp/midi/buffer.hpp>
 #include <pulp/midi/mpe_buffer.hpp>
+#include <pulp/midi/ump_buffer.hpp>
 #include <pulp/state/store.hpp>
 #include <string>
 #include <memory>
@@ -65,6 +66,15 @@ struct PluginDescriptor {
     /// The standard process() signature is unchanged; plugins that don't
     /// set this flag see no MPE-specific behaviour.
     bool supports_mpe = false;
+
+    /// Opt in to the native MIDI 2.0 UMP sidecar. When true, format
+    /// adapters that recognise UMP provide a UmpBuffer of full-resolution
+    /// channel-voice packets (16-bit velocity, per-note pitch bend,
+    /// per-note CCs) through Processor::ump_input() during process().
+    /// Adapters without native UMP transport synthesise the buffer by
+    /// converting the inbound MIDI 1.0 stream. `supports_mpe` and
+    /// `supports_ump` are independent and can both be set.
+    bool supports_ump = false;
 
     /// Tail time in samples (0 = no tail, -1 = infinite).
     /// Used by hosts to flush reverb/delay tails after playback stops.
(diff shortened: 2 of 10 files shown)


danielraffel:
Codex P1 (per-note CC index decoding) addressed in commit `5f835497`:

- `case 0x00` (Registered Per-Note Controller) now reads the controller index from byte 3 via `p.words[0] & 0x7F`, while the note number comes from `p.note_number()` (byte 2) — matching the MIDI 2.0 UMP per-note controller layout.
- Channel CC (`case 0xB0`) still decodes from byte 2 via `(p.words[0] >> 8) & 0x7F`, which is correct for channel voice messages (CC index is in byte 2 there).

Comment text rendered with the old (pre-fix) source because Codex snapshotted it against commit `3f06f2dd`; the current code already distinguishes the two layouts.

PR #142, opened 2026-04-12
skills(mpe): add MPE developer skill

danielraffel:
## Summary
- Adds `.agents/skills/mpe/SKILL.md` — three-step MPE synth scaffolding pattern, decision tree, four non-obvious gotchas (glide refcount on the steal path, base-class super-call requirement, zone RPN negotiation, pressure vs velocity), and current format-adapter coverage.

## Why
Closes the last open bullet in `planning/ralph-prompt-mpe.md`: skills were marked "create if a new repeatable workflow emerged". MPE synth creation is a multi-step pattern with several non-obvious gotchas worth codifying so future agents don't rediscover them.

## Test plan
- [x] Skill is documentation-only; no build/test impact
- [x] MPE test suite still 22/22 green locally

### DIFF
--- a/.agents/skills/mpe/SKILL.md
+++ b/.agents/skills/mpe/SKILL.md
@@ -1,0 +1,143 @@
+---
+name: mpe
+description: Build an MPE-aware Pulp synth — opt into MPE via PluginDescriptor, consume per-note pitch bend / pressure / timbre from MpeBuffer, and route voices through MpeVoiceAllocator without reinventing channel tracking.
+---
+
+# MPE
+
+Use this skill when adding per-note expression (pitch bend, pressure,
+CC 74 timbre) to a Pulp synth, or when writing a host that needs to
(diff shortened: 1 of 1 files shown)


danielraffel:
Both Codex findings addressed in 2454ed48:

- **P1 (step-3 code sample):** rewritten to use the real `Processor::process(output, input, midi_in, midi_out, ctx)` signature, `Processor::mpe_input()` accessor, per-event `allocator_.dispatch()` loop over `MpeBuffer::events()`, and index-based voice access via `allocator_.voice(i)` + `allocator_.polyphony()`. The snippet now compiles against the real APIs.
- **P2 (zone negotiation):** corrected — `MpeVoiceTracker::process()` does not parse RPN 6/7. The Gotchas section now says zones are configured via `MpeConfig` at tracker construction, not auto-negotiated, and points to `core/midi/include/pulp/midi/rpn_parser.hpp` for live RPN handling if needed.

PR #147, opened 2026-04-13
Fix Codex P1 + P2 review on PR #144 (versioning gates)

danielraffel:
Addresses the four Codex review comments on [PR #144](https://github.com/danielraffel/pulp/pull/144):

- **P1** — `already_bumped` returned True if ANY version file in a surface moved; multi-file surfaces (plugin.json + marketplace.json) could pass with only one bumped, causing split-brain versions.
  **Fix:** require ALL files to move. Report `✗ partial bump — not moved: <files>` and hard-fail. `apply_bumps` also switched to `all()`-semantics so a prior partial state self-heals on re-run.

- **P2** — Conventional-commit ceiling was computed from every commit in `base..head` without checking whether each commit's files touched the current surface. A plugin-only `feat:` could raise the SDK ceiling.
  **Fix:** loop now calls `git show --name-only` per commit and skips commits whose files don't match the surface's `trigger_paths`.

- **P2** — Skill-sync gate treated any file under `.agents/skills/<skill>/` as a SKILL.md update. Side files (notes, fixtures, logs) satisfied the gate spuriously.
  **Fix:** accept exact `SKILL.md` or nested `**/SKILL.md` only.

- **P2** — `auto-release.yml` idempotent-strict check compared `ls-remote --tags` output (tag object SHA for annotated tags) against `rev-parse HEAD` (commit SHA), so it never matched — tags would always fail-loudly on re-push.
  **Fix:** dereference via `<tag>^{}`; fall back to undereferenced SHA for lightweight tags.

## Test plan

- [x] `python3 tools/scripts/test_gates.py` — **13/13 green** including two new fixtures:
  - `test_partial_multi_file_bump_fails` (P1 regression guard)
  - `test_skill_side_file_does_not_satisfy_md_requirement` (P2 regression guard)
- [ ] CI: version-skill-check + build matrix

## Related
- Shipyard mirror fix: danielraffel/Shipyard (pending PR)

### DIFF
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -129,14 +129,25 @@ jobs:
               # this commit, skip. If it exists and points elsewhere, fail
               # loudly — a human must reconcile before the release binaries
               # can move.
+              #
+              # `git ls-remote --tags origin <tag>` returns the TAG OBJECT SHA
+              # for annotated tags (the default created by this workflow via
+              # `git tag -a`), not the commit SHA. Compare dereferenced
+              # target commits instead — the `^{}` suffix peels the tag
+              # object to its referenced commit.
               if git ls-remote --tags origin "$tag" | grep -q .; then
-                  remote_sha=$(git ls-remote --tags origin "$tag" | awk '{print $1}')
+                  remote_target=$(git ls-remote --tags origin "${tag}^{}" | awk '{print $1}')
+                  # Fallback: lightweight tag (no ^{} entry) — use the ls-remote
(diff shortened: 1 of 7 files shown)


danielraffel:
Codex P1 addressed in commit 4934c1e5: `apply_bumps` now derives the source version from the **base commit** rather than `v.current_version`. Refactored the version-extraction logic into `_extract_version_from_text` + `version_at_base` so partial-apply retries are idempotent — a half-bumped tree no longer double-bumps on the next run.

PR #172, opened 2026-04-14
Workstream 01 slice 1.7: MIDI CI ProfileReply handling

danielraffel:
## Summary

First workstream-01 slice. Fixes an unhandled message type and locks in the create_profile_inquiry layout with regression tests.

### Bug
`CiDiscovery::process_message` previously dispatched only `DiscoveryInquiry`, `DiscoveryReply`, and `ProfileInquiry`. `ProfileReply` hit `default: return {}` — profiles advertised by remote devices were silently dropped.

### Fix
Parses enabled/disabled profile lists from incoming ProfileReply SysEx and stores them per-source-MUID. Exposes:
- `remote_profiles(MUID)` getter
- `on_remote_profiles` callback

### Regression tests (test/test_midi_ci.cpp)
1. `create_profile_inquiry writes destination MUID` — locks in message layout (dest MUID at bytes 10..13). The production-readiness plan flagged this as a historical bug; the fix is already in `main`, this test prevents drift.
2. `parses ProfileReply and stores remote profiles` — full inquirer/responder round-trip with one enabled + one disabled profile; asserts inquirer sees both after processing the reply.

## Test plan
- [x] `ctest --test-dir build -R midi_ci --output-on-failure` → 10 test cases, 33 assertions, all pass
- [x] Library build clean

## Files
- `core/midi/include/pulp/midi/midi_ci.hpp` (+7 lines): `remote_profiles()`, `on_remote_profiles` cb, `remote_profiles_` map
- `core/midi/src/midi_ci.cpp` (+39 lines): ProfileReply parse + store
- `test/test_midi_ci.cpp` (+54 lines): two new regression tests

## Next in workstream 01
1.5 LV2 URID map feature resolution (real cross-host bug), then 1.1 CLAP multi-bus routing.

### DIFF
--- a/docs/status/support-matrix.yaml
+++ b/docs/status/support-matrix.yaml
@@ -70,6 +70,7 @@ scripted_ui:
     notes: Sibling theme.json overrides reload in ScriptedUiSession. Runtime-validated only in the standalone macOS scripted UI lane.
   state_preservation:
     status: usable
+    ci_test: pulp-test-state
     notes: Knob, fader, toggle, checkbox, toggle-button, and XY pad values are restored across scripted UI reloads. JavaScript heap state is rebuilt.
 
 audio_io:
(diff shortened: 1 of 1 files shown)


danielraffel:
Note: this branch was force-reset during the serial ralph loop. PR now carries workstream 08 sub-deliverable 8.4 (ci_test sweep batch 2): 5 additional `ci_test:` mappings in support-matrix.yaml, violation count drops 96→91 in report mode. The original MIDI CI ProfileReply slice survives in the parallel agent's local worktree and should be re-pushed under its own branch name. Title isn't editable via the CLI right now (GraphQL Projects-classic warning); the updated body below reflects actual contents.

PR #179, opened 2026-04-14
Workstream 01 slice 1.4: AUv3 MIDI sanitization

danielraffel:
## Summary

The AUv3 render block accepted any `AURenderEventMIDI` record and stuffed the first three bytes into a `choc::midi::ShortMessage` without validation, so SysEx, UMP, and packets with no status byte produced malformed short messages that downstream parsers would read as legitimate note/CC/pitchbend events.

## Fix
Construct the ShortMessage only when the packet is a well-formed short MIDI message: `length ∈ [1,3]` AND `data[0] & 0x80` (status byte). Longer packets are skipped cleanly until Pulp's `MidiEvent` gains sysex/UMP support (separate slice under audit 5.2 / workstream 01.6).

## Scope notes
This slice deliberately keeps scope tight. Two items from the full 1.4 deliverable land in follow-ups:
1. AUv3 sidechain bus — requires extending `_inputBusArray` with a second `AUAudioUnitBus` at init time
2. AUv3 MIDI-out via `MIDIOutputEventBlock` — plugin-side outbound path

## Test plan
- [x] `PulpGain_AU` builds clean; no new warnings on the touched code
- [ ] Host-smoke validation in AU Lab (AUv3) per planning acceptance

## Files
- `core/format/src/au_adapter.mm` (+17/-8 lines)

### DIFF
--- a/core/format/src/au_adapter.mm
+++ b/core/format/src/au_adapter.mm
@@ -370,18 +370,28 @@ struct AUBridge {
         }
 
         // MIDI events
+        // MIDI events. Pulp's MidiEvent is a choc::midi::ShortMessage
+        // (1–3 bytes); AU can also deliver longer messages (sysex, UMP)
+        // via AURenderEventMIDIEventList — we skip those here so a
+        // malformed ShortMessage is never constructed from truncated
+        // bytes. Workstream 01 slice 1.4.
         pulp::midi::MidiBuffer midi_in, midi_out;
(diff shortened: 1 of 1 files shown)


danielraffel:
Superseded: the MIDI sanitization landed as part of #183 (AUv3 sidechain + render-block routing). Closing to avoid conflict churn.

Issue #412, opened 2026-04-18
TSan hardening: suppressions, scoping, timeouts, audio-race hammer, path-gated triggers


danielraffel:
TSan is slow and not a real gate. Five improvements: 1) per-test timeout (landed in #401) 2) test/tsan.supp allow-list 3) scope to [audio|threading|events|host|state] tags 4) audio-race hammer test 5) path-gated trigger only on sync-primitive changes. See #401, #411.

danielraffel:
Closed by #401 + #431: all 5 TSan hardening items shipped — per-test timeout, tag scoping, tsan.supp, sync race-hammer, path-gated trigger. Audit evidence in 2026-04-20 housekeeping sweep.

Issue #411, opened 2026-04-18
ubsan: Toolbar add_separator() triggers BUS in ToolbarItem move-ctor (test #460)


danielraffel:
## Symptom

UBSan-instrumented build on macos-arm64 fails test #460 'Toolbar add items':

UndefinedBehaviorSanitizer:DEADLYSIGNAL
ERROR: UndefinedBehaviorSanitizer: BUS on unknown address
  #0  std::string::__default_init     string:1750
  #1  std::string::basic_string(string&&)  string:920
  #3  pulp::view::Toolbar::ToolbarItem::ToolbarItem(ToolbarItem&&)  app_framework.hpp:149
  #11 pulp::view::Toolbar::add_separator()  app_framework.cpp:181
  #12 CATCH2_INTERNAL_TEST_10()  test_gui_components.cpp:81

ASan + TSan pass; only UBSan catches it. macOS arm64 BUS = unaligned write. The crash is during vector::push_back reallocation moving an existing ToolbarItem (compiler-generated move ctor).

## Context

Surfaced by PR #401 (sanitizer-fix keystone). Pre-existing latent bug — the sanitizer matrix was failing on infra (private-submodule clone, GPU-target, lasso paren-init) before reaching test execution. #401 will skip test #460 in the UBSan job via --exclude-regex so the keystone can land. This issue tracks the actual fix.

## Likely fix path

ToolbarItem is a plain aggregate (3 strings, std::function, bool) — no custom move-ctor. libc++ __default_init BUS suggests an alignment issue with the parent Toolbar instance. Worth checking whether ToolbarItem is constructed at an unaligned address.

PR #424, opened 2026-04-18
ci: lift runner resolver, route sanitizers via repo vars (#412 step 6)

danielraffel:
## Move a job mid-incident without touching code

Lifts pulp's existing `resolve_runs_on()` runner-selection helper from inline Python in `.github/workflows/build.yml` into a reusable shared script, adds `local` as a third provider, and applies it to `sanitizers.yml`. Each sanitizer job (asan/tsan/ubsan/rtsan) is now independently routable between GitHub-hosted, Namespace, and local self-hosted runners via repo variables — no workflow edit, no full-matrix re-run.

Closes step 6 of #412 (issue references: #411 context, #401 recent sanitizer hardening).

## What changed

- `tools/scripts/resolve_runs_on.py` — new shared resolver with three modes (`provider`, `optional-namespace`, `default`). The first two preserve `build.yml` behavior exactly; `default` is new for per-target sanitizer routing.
- `tools/scripts/test_resolve_runs_on.py` — 14-case test suite covering build.yml parity, local-provider wiring, sanitizer default fallback, and JSON validation errors.
- `.github/workflows/build.yml` — replaces the inline Python heredoc with calls to the shared resolver. Zero behavior change: a 32-case behavioral diff against the previous inline logic passes with 0 mismatches.
- `.github/workflows/sanitizers.yml` — adds a `resolve-runners` job that outputs per-sanitizer `runs-on` values. Each sanitizer job reads its output via `fromJSON`. Every current hardening (per-test timeouts, TSan scope regex, concurrency group, push trigger, rtsan advisory gate, `submodules: false`) is preserved unchanged.
- `docs/guides/local-ci.md` — new section **"Switching a job's runner without a code change"** listing every `PULP_*_RUNS_ON_JSON` repo variable with `gh variable set` examples, plus a `gh runner register` appendix for self-hosted Mac setup.
- `.agents/skills/ci/SKILL.md` — short reference pointing ci operators at the new resolver + doc section (triggered by skill-sync gate because `.github/workflows/**` was touched).

## Load-bearing invariant (verified)

**Unset repo vars = today's behavior.** A repo that does nothing after this PR lands sees zero change:

- `build.yml` matrix with everything unset resolves to `macos-15` / `windows-latest` / `ubuntu-latest` on GitHub-hosted — identical to current.
- `sanitizers.yml` with everything unset resolves to `macos-14` (asan/tsan/ubsan) and `ubuntu-24.04` (rtsan) — identical to current.

Confirmed by end-to-end simulation of both `resolve-*` jobs plus a 32-case parity diff between the old inline build.yml resolver and the new shared resolver.

## Recommended first move

Set `PULP_SANITIZER_TSAN_RUNS_ON_JSON` to your local Mac runner labels:

```bash
gh variable set PULP_SANITIZER_TSAN_RUNS_ON_JSON \
  --body '["self-hosted","macos","arm64","sanitizer"]'
```

TSan benefits most from this. The GitHub-hosted `macos-14` runner is 3 vCPU / 7 GB; your local Mac is much bigger. Expect the scoped TSan sweep to drop from ~45 min to under 5 min. ASan / UBSan / RTSan can stay GitHub-hosted or move per preference — same mechanism, one variable per sanitizer. See `docs/guides/local-ci.md` § "Switching a job's runner without a code change" for the full variable list and a `gh runner register` appendix.

## Judgment call: shared script vs. composite action

Chose `tools/scripts/resolve_runs_on.py` because `tools/scripts/` is already the established home for repo-wide Python helpers (`skill_sync_check.py`, `version_bump_check.py`, `cli_sync_check.py`, `cmajor_external.py`, `jsfx_subset.py`). Workflows already invoke Python inline; swapping an inline heredoc for a shared script is a minimal delta. A composite action would still need Python for JSON validation and would fragment the test story across two dirs. The script is also trivially unit-testable (see `test_resolve_runs_on.py`).

## Out of scope (explicit, separate follow-up PRs)

- Wiring `local` as a provider for `build.yml`'s Linux and Windows legs. The resolver already understands `local`, and `PULP_LOCAL_{LINUX,WINDOWS}_RUNS_ON_JSON` are documented — but `build.yml` does not pass `--local-env` to the resolver for those legs in this PR. That is a small follow-up; it keeps this PR focused on sanitizers.
- Self-hosted runner registration on the Mac (ops task; commands only as docs appendix).
- #412 steps 2/4/5 (TSan suppressions, audio-race hammer, path-gated trigger) — unrelated to routing.
- #411 Toolbar UBSan fix — unrelated.

## Test plan

- [x] `python3 tools/scripts/test_resolve_runs_on.py` — 14/14 pass locally
- [x] Parity diff: 32 build.yml input combinations, 0 mismatches vs. old inline resolver
- [x] Simulated `resolve-runners` step: correct defaults when all vars unset; correct routing when `PULP_SANITIZER_TSAN_RUNS_ON_JSON` is set; `workflow_dispatch` input beats repo var
- [x] `skill_sync_check.py` passes (ci SKILL.md updated)
- [x] YAML parses via PyYAML on both workflows
- [ ] Shipyard cross-platform validation (this PR)

### DIFF
--- a/.agents/skills/ci/SKILL.md
+++ b/.agents/skills/ci/SKILL.md
@@ -182,6 +182,28 @@ macOS runs locally in parallel with Namespace Ubuntu/Windows.
 
 **Common mistake:** Pushing a branch and waiting for the auto-triggered GitHub Actions PR checks. Those use GitHub-hosted runners and are slow. Instead: cancel the auto-triggered run and dispatch on Namespace.
 
+### Runner selection is repo-variable driven (no code change needed)
+
+Every `runs-on:` decision in `.github/workflows/*.yml` routes through
+`tools/scripts/resolve_runs_on.py` and a per-target `PULP_*_RUNS_ON_JSON`
+repo variable. When every variable is unset the workflows resolve to
+their hard-coded defaults (backwards-compatible). Setting one variable
+moves one job — no workflow edit, no full-matrix re-run. The full
+variable list and `gh variable set` examples live in
+`docs/guides/local-ci.md` § "Switching a job's runner without a code
+change".
+
+Practical use case: if a sanitizer is slow on a GitHub-hosted runner
+(TSan on `macos-14` routinely exceeds 45 min), flip it to a self-hosted
+runner with `gh variable set PULP_SANITIZER_TSAN_RUNS_ON_JSON --body
+'["self-hosted","macos","arm64","sanitizer"]'`. The next sanitizer
+run picks up the new runner automatically.
+
+Sanitizer per-job `workflow_dispatch` inputs (`asan_runner_selector_json`
+/ `tsan_runner_selector_json` / `ubsan_runner_selector_json` /
+`rtsan_runner_selector_json`) are the equivalent one-off overrides for
+a single manual run.
+
 ```bash
 # Cancel auto-triggered GitHub-hosted run
 gh run cancel <run_id> --repo danielraffel/pulp
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,11 @@ jobs:
       provider: ${{ steps.resolve.outputs.provider }}
       matrix_json: ${{ steps.resolve.outputs.matrix_json }}
     steps:
+      - uses: actions/checkout@v5
+        with:
+          # Shared resolver lives in the repo at tools/scripts/resolve_runs_on.py.
+          # We only need the script; no LFS or submodules required.
+          lfs: false
       - id: resolve
         env:
           REQUESTED_PROVIDER: ${{ inputs.runner_provider || vars.PULP_DEFAULT_RUNNER_PROVIDER || 'github-hosted' }}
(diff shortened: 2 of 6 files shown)


danielraffel:
UBSan (and ASan/TSan) failures on this PR are an **env quirk on the self-hosted Mac runner**, not a Pulp bug.

Root cause in all three sanitizer jobs:

\`\`\`
##[error]Hook already exists: /Users/danielraffel/actions-runner/_work/pulp/pulp/.git/hooks/...
To resolve this, either:
  1: run \`git lfs update --manual\` for instructions on how to merge hooks.
  2: run \`git lfs update --force\` to overwrite your hook.
\`\`\`

\`actions/checkout@v5\` with \`lfs: true\` can't initialize LFS hooks because the runner's workspace has pre-existing customized git-lfs hooks from another checkout. Pulp's code never built.

## Fix

On the self-hosted Mac runner, once:

\`\`\`
cd /Users/danielraffel/actions-runner/_work/pulp/pulp
git lfs update --force
\`\`\`

That overwrites the conflicting hooks. Subsequent checkouts will pass.

## PR status

This PR's code is fine. The runner-selectable pattern and resolver are validated green in build.yml's jobs (which ran on Namespace, not local). Sanitizer-routing proof-of-concept works; we just can't see UBSan/ASan/TSan telemetry until the runner's LFS hook collision is cleared.

Recommend: fix the runner hook, retrigger CI, merge if green.

Issue #438, opened 2026-04-18
Post-merge Codex review sweep: 6 P1 + 5 P2 across today's cascade


danielraffel:
Per the \`feedback_codex_postmerge_audit\` practice, swept the Codex review
comments on the 12 PRs that merged today (#401, #416, #404, #407, #396,
#400, #425, #408, #406, #426, #392, #389). Found 11 items — filing here
as a single tracker so none fall on the floor.

### P1 (correctness / breakage risk)

- [x] **#416** — Include MSVC release gate in required Windows alias. The \`windows\` stable-name alias job depends on the build matrix but doesn't also gate on \`Windows MSVC release-path gate\`, so a green matrix + red release-path build will still report \`windows\` ✓. Widen the alias \`needs:\` to include the release-path job. — fixed via #439
- [x] **#407** — Map JNI memory levels to emitted Kotlin values. The JNI \`nativeOnMemoryPressure\` bridge in PulpActivity.kt passes raw Android \`TRIM_MEMORY_*\` constants (5/10/15/40/60/80) but the native side expects the Pulp enum (0/1/2). Map before forwarding. — fixed via #443
- [x] **#396** — Restrict macOS Android CLI fix to arm64 hosts. \`pulp doctor --fix\` downloads the macOS arm64 cmdline-tools bundle even on x86_64, silently installing a broken SDK. Gate on \`uname -m\`. — fixed via origin/main cli_common.cpp gating
- [x] **#396** — Preserve curl failures in Linux Android CLI fix command. The Linux path uses \`curl -s\` then \`unzip\` without \`set -e\` / \`pipefail\`; a 404 from Google silently continues into unzip and leaves a half-installed SDK. Use \`curl -fsSL\` + explicit check. — fixed via origin/main cli_common.cpp gating
- [x] **#425** — Clean partial clone target before retrying git clone. The \`retry_git\` wrapper retries \`git clone\` without removing the partial directory, so attempt 2+ fails with \"destination path already exists\" instead of actually retrying. — fixed via #441
- [x] **#389** — Treat optional Android CLI as non-failing check in \`pulp doctor\`. The \`android-cli\` check returns non-zero exit when the CLI is missing, but the CLI is optional — doctor should report it as a warning, not an error. — fixed via #442

### P2 (quality / edge cases)

[...]

PR #457, opened 2026-04-19
fix(midi): Win MIDI close() guards SysEx re-queue against shutdown race (#438 P1 / #388)

danielraffel:
## Summary

Codex review on #388 flagged that \`close()\` calls \`midiInReset\`, which
delivers queued SysEx buffers via \`MIM_LONGDATA\` — and the callback
unconditionally re-queues non-empty buffers via \`midiInAddBuffer\`.
If shutdown happens while a SysEx transfer is in flight, the re-queue
races with the unprepare/close sequence: headers left
\`MIDIERR_STILLPLAYING\` while \`close()\` clears backing vectors. Leak
or invalidated memory.

## What changed

- New \`closing_\` atomic. \`close()\` sets it true **before**
  \`midiInStop\`/\`midiInReset\` and clears after \`midiInClose\`.
- \`MIM_LONGDATA\` and \`MIM_LONGERROR\` callback paths skip
  \`midiInAddBuffer\` while \`closing_\` is true. Same pattern as the
  Linux env teardown fix in #449.

## Test plan

- [x] Compiles
- [ ] Manual: stress-test open/close on Windows with active SysEx
  traffic; previously left STILLPLAYING headers, now clean

## Relates

- Closes the #438 P1 item for #388

### DIFF
--- a/core/midi/platform/win/winmidi_device.cpp
+++ b/core/midi/platform/win/winmidi_device.cpp
@@ -102,6 +102,15 @@ public:
 
     void close() override {
         if (handle_) {
+            // Set the closing flag BEFORE midiInReset so any
+            // MIM_LONGDATA callback fired during reset (or already in
+            // flight when stop() races with an arriving SysEx) skips
+            // the midiInAddBuffer re-queue. Without this, a re-queued
+            // header collides with the unprepare/close sequence and
+            // either leaks (MIDIERR_STILLPLAYING) or invalidates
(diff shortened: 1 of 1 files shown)


danielraffel:
Retrying CI — failure was an unrelated SPSC race-hammer flake (`SpscQueue hammer: producer + consumer for 500ms` — produced=7196582 consumed=7196581, classic last-element timing window). PR change is in winmidi only.

Issue #413, opened 2026-04-18
dx: plugin isolation guard + feature-request volunteering (agent-assisted)


danielraffel:
Part A — plugin isolation guard: pre-commit hook flagging core/examples diffs from plugin branches; doctor check for ambiguous repo layouts; docs page on plugin-vs-framework code boundary.

Part B — feature-request volunteering: pulp feature-request CLI + /pulp-suggest-feature slash command. Three modes: request-only, request+code-donation (per-submission consent), or donate-prompt-only (Claude plugin helps harden the prompt first). Ships with a .github/ISSUE_TEMPLATE/feature-request.yml so the form works standalone too.

danielraffel:
## Scope expansion — add Part C: bug report template + anonymized diagnostic donation

Since these will be Pulp's **first** GitHub issue templates, bug reports belong in the same system as feature requests. Same design principles: CLI-assisted, user owns what gets sent, standalone fill-in works too.

## Part C — Bug report template + diagnostic donation

### Standalone template

`.github/ISSUE_TEMPLATE/bug-report.yml` with sections:
- Summary (1-line)
- Pulp version (auto-detect hint: `pulp version`)
- Platform + OS version
- Plugin format being built (VST3 / AU / CLAP / AUv3 / Standalone / other)
- DAW / host if applicable, with version
- Steps to reproduce
- Expected vs actual behavior
- Logs / stack trace if available
- **Optional** attachment section: "run `pulp bug-report --attach` to generate a bundle; you'll preview + edit before it uploads"

### CLI diagnostic bundle

`pulp bug-report` — interactive. Generates a sharable bundle at a temp path the user can inspect and edit.

**What goes in:**
- `pulp version` output
- `pulp doctor` output (minus any environment-variable dumps that might contain secrets)
- Last N lines of `~/Library/Logs/Pulp/*.log` (or platform equivalent)
- CMake cache snapshot (no absolute paths — normalized to `$PULP_ROOT`)
- Build compiler + flags
- Crash log if user points the CLI at one (`--crash-log <path>`)
- Generated "plugin fingerprint": targets configured, formats enabled, GPU backend, platform deps

**Anonymization pass** — applied BEFORE writing the bundle:
- Home directory → `$HOME`
- Username → `$USER`
- Computer name → `$HOSTNAME`
- IP addresses (IPv4 + IPv6) → redacted
- MAC addresses → redacted
- Any email-shaped strings → redacted
[...]

PR #458, opened 2026-04-19
docs: defer pulp report bug guidance until CLI ships (#438 P2 / #435)

danielraffel:
## Summary

Codex review on #435 flagged that the bug template told filers to run
\`pulp report bug\` / \`pulp report bug --attach\`, but the \`report\`
subcommand isn't shipped yet (no entry in \`tools/cli/pulp_cli.cpp\`).
Following the instruction yields an unknown-subcommand error — the
opposite of the friction reduction the template should provide.

## What changed

- Replace \`pulp report bug\` instructions with \`pulp doctor\` output
  guidance (which exists and is useful).
- Note that the automated diagnostic-bundle CLI is roadmap-tracked
  in #413; template will point at it once it lands.

## Relates

- Closes the #438 P2 item for #435

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -7,11 +7,11 @@ body:
     attributes:
       value: >
         Thanks for filing a bug. The more specific the repro, the faster
-        someone can land a fix. If you'd like Pulp to collect a diagnostic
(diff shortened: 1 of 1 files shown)


danielraffel:
Retrying CI — failure was an unrelated flake on `pulp doctor android|ios are recognized subcommands` (timeout at 32s on macOS ARM64; PR is docs-only).

Issue #500, opened 2026-04-19
post-merge Codex sweep (2026-04-20): 6 P1 + 4 P2 findings on today's cascade


danielraffel:
## Context

Post-merge Codex review of today's cascade (#477, #478, #480, #482, #483, #484, #485) surfaced new actionable findings after our earlier sweep (#488-#493). These are NEW (not duplicates of the earlier sweep).

Sibling to #438 pattern — grouping for triage; each sub-item becomes its own PR.

## P1 (ship soon)

- [ ] **#482** — \`core/audio/src/audio_focus.cpp:52\` — \`AudioFocusRegistry::publish()\` snapshots callbacks, then invokes them AFTER a Token::reset (from within a callback) has mutated \`cbs_\`. Same class of use-after-free that \`pulp::platform::Environment\` already handles by copying-under-lock-and-dispatching-outside-lock. The copy-outside-dispatch pattern is there but the ordering is wrong. Fix: invoke via the pre-snapshot copy, not the post-mutation vector.

- [ ] **#482** — \`core/platform/src/android/jni_bridge.cpp:305\` — JNI audio-focus path maps BOTH \`AUDIOFOCUS_LOSS_TRANSIENT\` and \`AUDIOFOCUS_LOSS_TRANSIENT_CAN_DUCK\` to \`duck\`. Listeners can never observe \`lost_transient\`; the enum has 4 states but the bridge only produces 3.

- [ ] **#483** — \`android/app/src/main/kotlin/com/pulp/audio/PulpAudioController.kt:110\` — fast start/stop race: \`stopAudio()\` skips \`unbindService\` when \`bound == false\`, but \`onServiceConnected\` can flip \`bound = true\` asynchronously AFTER stopAudio returns, leaving a dangling binding. Fix: always call unbindService in a try/catch regardless of \`bound\` flag (matches Android framework guidance).

- [ ] **#485** — \`core/view/platform/win/accessibility_win.cpp:212\` — \`PulpHostProvider\` holds a raw \`UiaSession*\`. \`shutdown_accessibility\` does \`host_provider->Release()\` then \`delete session\`. If UIA clients still hold a ref to the provider after our Release (they cache references), the next method call on the provider dereferences freed \`session_\`. UAF. Fix: hold the session by shared_ptr so the provider's Release path destructs the session only when the last UIA ref drops.

[...]

PR #506, opened 2026-04-20
view(win): UIA shutdown UAF — disconnect provider before delete (#500)

danielraffel:
## Summary

`PulpHostProvider` holds a raw `UiaSession*`, and `shutdown_accessibility` called `host_provider->Release()` then `delete session`. Screen readers cache provider pointers, so any in-flight method call reached the freed `session_` → UAF, crash risk under re-attach.

Fix: call `UiaDisconnectProvider(host_provider)` before `Release()`. This is the Windows-documented shutdown barrier — it waits for in-flight UIA calls to drain and rejects subsequent calls. After it returns, no UIA-originated call can reach the provider, so session deletion is safe.

Flagged by Codex as part of the #485 review, rolled up under #500.

## Test plan

- [x] Extended `test_accessibility_provider.cpp` with a 32-cycle init+shutdown hammer test that exercises the new barrier path
- [x] Existing 2-cycle test kept
- [ ] CI green on macOS / Ubuntu / Windows (Windows path is the one that actually drives the real UIA barrier)

Refs #500, #485.

### DIFF
--- a/.agents/skills/ci/SKILL.md
+++ b/.agents/skills/ci/SKILL.md
@@ -50,6 +50,12 @@ period (see danielraffel/pulp#120).
 # Primary: Shipyard
 shipyard run                              # validate current branch
 shipyard ship                             # PR + validate + merge on green
+shipyard ship --resume                    # pick up an interrupted ship (v0.3.0+)
+shipyard ship --no-resume                 # discard stale state, start fresh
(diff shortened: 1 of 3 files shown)


danielraffel:
Wrong head ref — re-opening from the correct worktree branch.

PR #486, opened 2026-04-19
ci(android): re-enable emulator smoke test (#14 + #33)

danielraffel:
Closes #14 (generated Android app end-to-end smoke, native-init half) and the Android half of #33 (mobile emulator CI smoke).

Replaces the disabled hand-rolled avdmanager/emulator harness with reactivecircus/android-emulator-runner@v2 on ubuntu-latest + KVM. Runs x86_64 API 34 google_apis image with `-gpu swiftshader_indirect` — no real GPU needed.

Scope is deliberately minimal:
1. Install APK from android-build artifact
2. Launch PulpActivity
3. Tail logcat for 'JNI_OnLoad: Pulp native bridge initialized' (fires from System.loadLibrary in PulpApplication.onCreate — BEFORE rendering or foreground service startup)
4. 5-second hold + process-alive check (catches secondary-init crashes)

Pass = init log + process alive. Diagnostic logs uploaded on every run. Full rendering-path smoke still waits on dedicated GPU-capable emulator infra — this job is the 'native bridge didn't regress' signal, cheap enough for every PR.

PR #530, opened 2026-04-20
ci(android): re-enable emulator smoke test with correct ABI + intent

### LINKED ISSUE #487: ci: document Android emulator CI installation requirements (ubuntu-latest)
## Context

PR #486 (\`ci(android): re-enable emulator smoke test\`) hit a known-but-undocumented failure pattern on first run: \`ERROR | Unable to connect to adb daemon on port: 5037\` from \`reactivecircus/android-emulator-runner@v2\`. The fix requires a specific set of preparatory steps on the CI runner that are NOT the same as running an emulator locally via Android Studio.

This issue exists to document those requirements + the failure modes we've hit, so future contributors don't have to rediscover them.

## Why CI is different from a local dev machine

A developer's Mac running Android Studio has, out of the box:

- Android Studio IDE + all cmdline-tools
- Full Android SDK at \`~/Library/Android/sdk\`
- Hardware virtualisation already configured (HAXM on Intel, Apple Virtualisation Framework on Apple silicon)
- System images pre-downloaded from past AVD creations
- \`adb\` + \`sdkmanager\` + \`avdmanager\` on PATH via the IDE's env

\`ubuntu-latest\` GitHub runners have **none** of this pre-installed. Every run starts from a bare VM and must install everything before the emulator can boot. The \`reactivecircus/android-emulator-runner\` action tries to do this automatically, but on recent ubuntu images the auto-install path is unreliable.

## Required CI steps (documented in \`.github/workflows/android.yml\`)

1. **Enable KVM group perms** — ubuntu GitHub runners have \`/dev/kvm\` but not world-readable by default. Without this step the action boots a software emulator that takes 15+ minutes.

   \`\`\`yaml
   - name: Enable KVM group perms
     run: |
       echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
       sudo udevadm control --reload-rules
       sudo udevadm trigger --name-match=kvm
   \`\`\`

2. **Install Java 17** — the emulator action requires it (\`actions/setup-java@v4\` with \`temurin\`/\`17\`).

3. **Install Android SDK explicitly BEFORE the emulator action** — \`android-actions/setup-android@v3\`. This is the fix for the \`adb daemon connection failure\` mentioned above. The emulator-runner action assumes \`sdkmanager\` is on PATH; without this step it isn't, and the whole job silently fails mid-setup.

4. **Use the emulator-runner action with explicit boot timeout**. The default 600-second budget is tight for API 34; API 30 + 900s boot timeout is the action maintainer's recommended combo.

5. **\`-gpu swiftshader_indirect\`** in emulator-options — no real GPU on the runner. Software rendering is fine for native-init smoke (which is all this job does). Full rendering-path smoke still needs dedicated GPU-capable emulator infra (see #77).

## Developer-side impact

**None** for running Pulp locally — your existing Android Studio install is untouched and works as normal for local development. The \`android/app/build.gradle.kts\` + \`PulpAndroid.cmake\` paths are unchanged.

This issue + the workflow changes only affect **what the GitHub CI runner does** on each Android PR.

## Follow-ups / sibling issues

- **#77** — mobile emulator/simulator CI smoke (umbrella tracker; this workflow closes the Android native-init half; iOS simulator smoke + full-rendering emulator path still open)
- **#337** — generated Android app E2E smoke (this workflow closes the native-init half; full-app smoke waits on GPU-capable emulator infra)
- **#14** — closed by #486 once it merges

## Acceptance criteria

- [x] \`.github/workflows/android.yml\` documents the KVM + setup-java + setup-android + emulator-runner sequence inline (landed in #486)
- [ ] PR #486 merges green (retrigger pending as of this issue)
- [ ] \`docs/guides/android.md\` or \`.agents/skills/android/SKILL.md\` gets a 3-sentence "how Android CI works" section pointing at the workflow and this issue
- [ ] Future CI failures on the Android path route through this issue for triage

## Labels
\`ci\`, \`android\`, \`documentation\`

danielraffel:
## Summary

Re-enables the `android-emulator-test` CI job in `.github/workflows/android.yml`. The job was gated on an unsettable repo variable (`PULP_ANDROID_EMULATOR_ENABLED`) after PR #486 hit what looked like an emulator boot failure three times. Post-merge Codex review on #487 found the real cause wasn't the emulator at all — the smoke script itself was broken.

## Root cause (from #487)

1. **ABI mismatch.** `android/app/build.gradle.kts` pins `abiFilters = listOf("arm64-v8a")` only (GPU/Skia pre-built for arm64). Earlier iterations of the workflow targeted an `x86_64` system image, so `adb install` returned `INSTALL_FAILED_NO_MATCHING_ABIS` and every subsequent adb command looked like it hung.
2. **Wrong activity intent.** The APK's `applicationId` is `com.pulp.app`; an earlier `am start -n com.pulp/com.pulp.PulpActivity` invocation would not resolve.

Both the AVD arch (`system-images;android-34;google_apis;arm64-v8a`) and the launch intent (`com.pulp.app/com.pulp.PulpActivity`) are already correct in the current workflow — the only remaining blocker was the gate. This PR removes it and adds a comment block above the job explaining why AVD arch + launch intent must stay matched to `build.gradle.kts`.

## Why this is safe to re-enable unconditionally

- The job only runs when `android/**`, `core/**`, `tools/cmake/PulpAndroid.cmake`, or this workflow changes (path filters at the top of the file).
- `timeout-minutes: 20` caps the runtime, so a still-broken run burns 20 min of budget, not the whole CI lane.
- It `needs: android-build`, so it cannot start until the APK + `libpulp.so` verification has passed on macOS.

## Test plan

- [x] yamllint + actionlint clean
- [x] Structural YAML parse round-trips
- [x] skill_sync_check.py + version_bump_check.py pass in --mode=report
- [ ] Android Build job (macOS + Windows) stays green
- [ ] `android-emulator-test` actually boots the arm64-v8a emulator on `macos-latest`, installs the APK, launches `PulpActivity`, and greps `JNI_OnLoad: Pulp native bridge initialized` out of logcat

Closes #487.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -106,15 +106,19 @@ jobs:
           path: android/app/build/outputs/apk/debug/app-debug.apk
           retention-days: 7
 
-  # Emulator test — only on macOS (KVM not available on GitHub Windows runners)
-  # Disabled until emulator CI infrastructure is set up (see #77).
-  # The emulator test requires GPU support (Dawn/Vulkan) which CI runners
-  # don't have, causing a guaranteed 20-minute timeout on every PR.
-  # actionlint flags `if: false` as a constant expression; we gate on an
-  # unsettable GitHub expression instead so the job is skipped at plan
-  # time without triggering the [if-cond] lint.
+  # Emulator test — only on macOS (KVM not available on GitHub Windows runners).
+  # This is a native-init smoke: boots a headless arm64-v8a emulator, installs
+  # the arm64-only debug APK, launches PulpActivity, and greps logcat for the
+  # JNI_OnLoad marker. Full GPU/rendering validation is out of scope here
+  # (tracked under #77).
+  #
+  # Re-enabled in #487 after identifying the real cause of earlier failures:
+  # the workflow was targeting an x86_64 system image while the APK only
+  # contains arm64-v8a (see android/app/build.gradle.kts), which produced
+  # INSTALL_FAILED_NO_MATCHING_ABIS and made every subsequent adb command
+  # appear to hang. The AVD arch and the `am start` intent below are the
+  # matched pair — touch both together if the applicationId ever changes.
   android-emulator-test:
-    if: ${{ vars.PULP_ANDROID_EMULATOR_ENABLED == 'true' }}
     runs-on: macos-latest
     needs: android-build
     timeout-minutes: 20


danielraffel:
## Run 1 failure analysis

The `android-emulator-test` job failed after the 20-minute `timeout-minutes` cap. Root cause (from the job log):

```
HVF error: HV_UNSUPPORTED
WARNING      | QEMU main loop exits abnormally with code 1
qemu-system-aarch64-headless: failed to initialize HVF: Invalid argument
```

`github.com/danielraffel/pulp/actions/runs/24655977992/job/72091082998`

This is a third failure mode beyond the two identified in #487:

1. ~~ABI mismatch (x86_64 emulator vs arm64-only APK)~~ - already fixed
2. ~~Wrong activity intent~~ - already fixed  
3. **NEW: `macos-latest` is ARM64 (Apple Silicon), and HVF is rejecting the arm64-v8a system image with `HV_UNSUPPORTED`.** The emulator binary starts, then QEMU immediately exits with status 1 after `failed to initialize HVF: Invalid argument`. `adb wait-for-device` hangs, and the job burns its 20-minute budget.

**Why this is a conflict, not a config fix:**

- The APK is `arm64-v8a` only (`abiFilters = listOf("arm64-v8a")` in `android/app/build.gradle.kts`) because GPU/Skia is pre-built for arm64 only.
- On a macOS **ARM64** runner, `qemu-system-aarch64-headless` can't use HVF to accelerate an arm64 Android guest — this was historically the HVF path that worked on Intel macOS runners with x86_64 guests, but Apple's HVF on Apple Silicon hosting arm64 Android via QEMU is broken/unsupported as of API 34 emulator images (android-34 google_apis/arm64-v8a).
- Switching to x86_64 emulator + x86_64 APK would work, but requires producing an x86_64 `libpulp.so` (Skia prebuild doesn't currently have x86_64 Android binaries in `external/skia-build`).

**Paths forward (any one unblocks this):**

1. **Add x86_64 to debug-build `abiFilters`** and accept that GPU paths are stubbed/disabled on x86_64 Android — the emulator smoke only validates `JNI_OnLoad`, not rendering. Requires a Skia x86_64 Android binary or a CMake gate that disables Skia for x86_64 builds.
2. **Run the emulator on a Linux arm64 runner via KVM** (Namespace has `ubuntu-24.04-arm64` available). KVM on ARM hosts running arm64 guests works natively — no HVF dependency.
3. **Skip emulator in CI** and rely on the `android-build` artifact verification (libpulp.so present in APK) plus manual on-device smoke before releases. This is what was effectively in place before this PR.

I am NOT forcing this PR to merge. The CI workflow change is correct (re-enables the job with correct AVD + launch intent), but the underlying HVF issue on `macos-latest` is a separate problem not solvable in this PR.

Leaving this open for the issue author to decide which path to take. Closing #487 as "root-cause identified" is reasonable — the original two causes are fixed; the third is now documented here.

PR #708, opened 2026-04-24
fix(ci): MSVC /utf-8 for UTF-8 argv on Namespace Windows (#707)

### LINKED ISSUE #707: ci(windows): test binaries need `/utf-8` MSVC compile flag so argv decodes UTF-8 on Namespace
## Problem

On Namespace Windows runners, \`ctest\` filters with em-dashes in test names (e.g. \`RenderLoop factory returns a loop that starts and stops — issue-646\`) get mangled to \`G??\` in the test binary's argv — 46+ spurious test-filter-no-match failures per run. Started failing #702 and will fail every subsequent PR with UTF-8 characters in test names.

Observed on runs 24872275599, 24873833616, 24874759281 (all PR #702 retries).

## What I tried (and why it didn't work)

**PR #697** — \`chcp.com 65001\` in bash shell before ctest. **Worked on GitHub-hosted Windows, failed on Namespace Windows.** Different shell/subprocess pipelining.

**PR #702 follow-up** — wrapped ctest in \`shell: cmd\` with \`chcp 65001 && ctest ...\`. Log confirms \`Active code page: 65001\`, but em-dash STILL mangles. Console code page changes OUTPUT encoding, not argv INPUT encoding.

## Root cause

The Windows CRT's \`argv\` is constructed from \`GetCommandLineW()\` (UTF-16) then narrowed via \`WideCharToMultiByte(CP_ACP, ...)\`. \`CP_ACP\` is the **system ANSI code page**, set at OS install time (typically CP-1252 on US Windows). It's NOT the console code page set by \`chcp\`.

When \`ctest\` spawns a test binary with a filter arg containing \`—\` (U+2014), the spawn system passes UTF-16 → the child's CRT converts to narrow via CP_ACP → \`—\` becomes \`??\` because CP-1252 doesn't have that glyph.

GitHub-hosted Windows runners happen to work because their image may have \`CP_ACP=65001\` set (or use a different spawn pipeline that preserves UTF-16 argv via \`wmain\`). Namespace Windows does not.

## Fix

**Compile test binaries with MSVC \`/utf-8\` flag.** This makes the CRT use UTF-8 for all narrow-string operations including \`argv\` — regardless of console CP or system ACP.

Add to \`tools/cmake/*.cmake\` or the test CMakeLists:
\`\`\`cmake
if(MSVC)
    target_compile_options(<test-target> PRIVATE /utf-8)
endif()
\`\`\`

Or, better, set it as a Pulp-wide compile option so CHOC / Catch2 / test code all see consistent UTF-8:
\`\`\`cmake
if(MSVC)
    add_compile_options(/utf-8)
endif()
\`\`\`

## Alternative (heavier)

App manifest declaring UTF-8 ACP (Windows 10 1903+ only). Requires per-target manifest files. Not recommended.

## Related

- [Shipyard #208](https://github.com/danielraffel/Shipyard/issues/208) — Shipyard-side UTF-8 prelude for ssh-windows (closed in v0.38.0; covers commands Shipyard dispatches, not GH Actions).
- PRs #697, #699, #702 — history of the workaround attempts on the workflow side.
- #683 — Namespace storage migration (where this surfaced first).

## Urgency

P1 — blocks every PR's Windows (x64) [namespace] lane until fixed. Currently admin-merging through it.

danielraffel:
## Summary

Closes [#707](https://github.com/danielraffel/pulp/issues/707). Adds \`/utf-8\` to MSVC \`add_compile_options\` in the top-level CMakeLists and reverts the complex Windows-specific cmd-shell Test step in build.yml back to the simple single ctest line.

## Root cause (validated)

MSVC's CRT narrows wide argv via \`CP_ACP\` (system ANSI code page), not the console code page \`chcp\` affects. Namespace Windows has CP_ACP=CP-1252; every PR's Windows (x64) [namespace] lane today fails 46+ em-dash test cases because test filters get mangled from \`—\` to \`G??\` before Catch2 can match them.

Today's workaround attempts on the workflow side (#697 bash chcp, #702 cmd-shell chcp wrapping) changed console CP but never CP_ACP, so the failure persisted.

\`/utf-8\` makes MSVC's CRT do UTF-8 narrowing regardless of CP_ACP. It's the actual fix at the right layer.

## Changes

- `CMakeLists.txt` — add `/utf-8` to the MSVC `add_compile_options` block with a reference comment to #707.
- `.github/workflows/build.yml` — revert the split `Test (non-Windows)` / `Test (Windows — UTF-8 code page wrapped)` steps from #702 back to the single `Test` step. `/utf-8` obviates the cmd-shell workaround.

## Expected impact

- Windows (x64) [namespace] goes green on em-dash tests without admin-merge.
- Unblocks the admin-merge-on-Windows pattern we've been stuck in since 4/23.
- No runtime behavior change on non-Windows builds.

## Next

If this lands clean, close #707 after the first normally-merged PR where Windows (x64) [namespace] turns green.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -239,30 +239,10 @@ jobs:
         run: ccache --show-stats
         shell: bash
 
-      - name: Test (non-Windows)
-        if: runner.os != 'Windows'
+      - name: Test
         working-directory: build
-        shell: bash
         run: ctest --output-on-failure -C Release -LE validation -j4 --exclude-regex "STFT|VisualizationBridge|Clipboard|Toolbar add items"
-
-      - name: Test (Windows — UTF-8 code page wrapped)
-        # Windows needs a UTF-8 console code page so ctest's child
-        # processes receive filter args (test names with em-dashes
-        # like "applies_if — empty expression") without CP-1252
-        # mangling. Wrapping ctest inside `cmd /c "chcp 65001 & ctest"`
-        # ensures the code page applies to the cmd subshell that
-        # spawns test binaries — a plain bash `chcp.com 65001` ran
-        # at the outer shell level but didn't propagate to ctest's
-        # children on Namespace Windows (observed on #702).
-        # Shipyard v0.38.0's UTF-8 prelude only covers commands
-        # Shipyard dispatches via ssh-windows; GH Actions ctest
-        # calls don't go through that path.
-        if: runner.os == 'Windows'
-        working-directory: build
-        shell: cmd
-        run: |
-          chcp 65001
-          ctest --output-on-failure -C Release -LE validation -j4 --exclude-regex "STFT|VisualizationBridge|Clipboard|Toolbar add items"
+        shell: bash
 
       - name: Upload plugin bundles (macOS)
         if: success() && runner.os == 'macOS'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,7 +69,16 @@ set(PULP_ARA_SDK_DIR "" CACHE PATH "Path to a developer-supplied Celemony ARA SD
 
 # ── Compiler Warnings ──────────────────────────────────────────────────────
 if(MSVC)
-    add_compile_options(/W4 /wd4100)
+    # /utf-8 makes MSVC treat source files AND narrow string literals as
+    # UTF-8, and makes argv decode via UTF-8 regardless of the system
+    # ANSI code page (CP_ACP). Without this, test names like
+    # "applies_if — empty expression" get mangled to "G??" when ctest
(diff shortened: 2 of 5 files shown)


danielraffel:
Superseded by simpler fix: [#TBD — fix/ascii-hyphens-in-test-names](https://github.com/danielraffel/pulp/tree/fix/ascii-hyphens-in-test-names). The `/utf-8` + workflow wrapper approach here was over-engineered; pure ASCII hyphens in TEST_CASE names sidesteps the whole encoding stack (CP_ACP, chcp, catch_discover_tests pipe). Codifying that as the convention going forward.

PR #735, opened 2026-04-24
feat: harden project SDK bump flow

danielraffel:
Teach pulp project bump to treat standalone pulp.toml sdk_version as the source of truth, keep versioned find_package(Pulp ...) in lockstep, and leave project(... VERSION ...) as the app/plugin product version.

Add safety rails for source-checkout refusal, CLI skew, redundant origin/main pins, dirty pin files, and managed-vs-custom sdk_path handling. Share standalone SDK resolution with build and doctor so stale sdk_path values are surfaced before configure work becomes misleading.

Test Plan:

- cmake --build build --target pulp-cli pulp-test-cli-project-bump pulp-test-cli-shellout -j10

- ./build/test/pulp-test-cli-project-bump "[project-bump]"

- ./build/test/pulp-test-cli-shellout "[issue-244]"

- ./build/test/pulp-test-cli-shellout "pulp project is a recognized command with bump + undo subcommands"

- ./build/test/pulp-test-cli-shellout "pulp project bump rejects non-semver --to"

- python3 tools/scripts/cli_sync_check.py

- git diff --check

Version-Bump: cli=minor reason="add standalone project SDK bump safety"


danielraffel:
Closing the two remaining standalone bump review threads here via follow-up #770 (659892ec). That follow-up now skips no-op managed sdk_path rewrites when the project is already at the target SDK, and it still repairs standalone find_package(Pulp ...) drift even when origin/main already pins the target SDK.

PR #782, opened 2026-04-25
test(format): cover VST3 adapter process path

danielraffel:


danielraffel:
Addressed the Codex review note in `0fd5f83c`: the VST3 process-context test now sets `kTempoValid` when asserting tempo propagation. Local `pulp-test-vst3-plugin-state` still passes (`118` assertions / `5` cases).

danielraffel:
The current TSan failure is checkout infrastructure, not a sanitizer/test failure: the job failed during Git fetch with `curl 18`, `unexpected disconnect`, and `bad pack header` before tests ran. GitHub will not rerun failed jobs while the sanitizer workflow still has queued jobs; I will rerun failed jobs after run `24929565716` reaches a terminal state.

danielraffel:
Reran failed jobs for Sanitizer Tests run `24929565716` now that the workflow is terminal. The only failed job was the earlier checkout partial-pack/RPC disconnect before tests.

Issue #967, opened 2026-04-29
view: View widget default background should be transparent, not opaque white


danielraffel:
Tracked under #924. Spectr#28 native-bridge. Strongly related to #964.

## Problem

`View` widget defaults to **opaque white** background when no `setBackground` call is made. Web/HTML semantics: `<div>` defaults to **transparent**. This mismatch causes any layered HTML/React port to look broken until every container explicitly sets `background: transparent`.

## Concrete impact

Spectr's FilterBank wrap (`<div style={{position:'absolute', inset:0}}>`) and overlay canvases inherit white default — covering the actual app background and breaking the visual layering the bundle assumes.

## Reproduction

Diagnostic: setting `background: 'rgba(0,200,0,0.4)'` on the FilterBank wrap shows green is visible behind translucent toolbars but the center area shows plain white where canvas children sit (their View parent paints opaque white before the canvas surface paints).

## Suggested fix path

In `core/view/src/view.cpp` `paint_self_and_children`:

```cpp
// before
if (background_color_set_) canvas.fill_rect(bounds, background_color_);
else canvas.fill_rect(bounds, kDefaultWhite);  // ← problem

// after
if (background_color_set_) canvas.fill_rect(bounds, background_color_);
// no else — transparent default matches HTML/CSS
```

For users who DO want an opaque white container, they explicitly call `setBackground(id, '#fff')` — same one-line call most existing Pulp examples already make for legible UI on a transparent surface.

## Risk

This changes the default behavior for every existing widget. Audit:
- Plugins that intentionally rely on white default → will now show through to the host's window-bg
- Standalone host renders → will need an explicit white root background, which the standalone shell can do once

Mitigation: the standalone wrapper / plugin-host wrapper sets root background to white explicitly. All existing plugins built against current Pulp will still look right because the outermost root is opaque.

## Acceptance criteria

- [ ] `View::paint_self_and_children` no longer fills opaque white when bg unset
- [ ] Existing `core/view/test/test_view.cpp` updated for new default
- [ ] Standalone host wrapper sets root bg explicitly so default plugins still look like they did
- [ ] Spectr's FilterBank center is visible without explicit `background: transparent` workarounds (#964 closes too)

danielraffel:
**Closing as premise-stale — no fix needed.**

[...]

Issue #964, opened 2026-04-28
view: FilterBank canvas2D draw output not reaching visible surface (was: child Views opaque-white over canvas)


danielraffel:
Tracked under #924. Surfaced while debugging Spectr's FilterBank.

## Symptom

Green diagnostic background applied to the FilterBank wrap (`<div style={{position:'absolute', inset:0, background:'rgba(0,200,0,0.4)'}}>`) is **only** visible:
- in the strip directly above the bottom toolbar
- behind the bottom toolbar (translucent)

It is NOT visible in the center of the wrap (where it should show through transparent canvas children).

## Tree

```
App (pr_2f)
├── FilterBank wrap (pr_3) — green bg via setBackground
│   ├── canvas pr_1 — main FilterBank canvas (transparent default per #929)
│   └── canvas pr_2 — overlay canvas (transparent default per #929)
├── top toolbar (pr_10) — translucent gradient
├── StatusBanner (pr_11) — opacity 0 by default; 19 props applied
└── bottom toolbar (pr_2e) — translucent gradient
```

## Expected

Transparent canvas children + transparent StatusBanner (opacity 0) → green wrap shows through entire center area.

## Actual

Center is opaque white. Either:
1. Canvas widgets aren't fully transparent post-#929 / #934 / #955
2. StatusBanner with opacity:0 still composites opaquely
3. Some other compositing pass paints opaque white between the wrap and its children

## Reproduction

Spectr branch `feature/native-react-editor` at SDK 0.60.0:

```bash
cd ~/Code/spectr
# (locally apply: green bg on FilterBank wrap)
cd native-react && npm run build:port
cd ../build && cmake --build . --target Spectr_Standalone -j$(sysctl -n hw.ncpu)
./native-react/tools/screencap-spectr.sh native /tmp/test.png
```

Screenshot: pure green is visible ONLY at bottom strip + behind bottom toolbar; center area is opaque white covering the green wrap.

## Notes

- `bridge-coverage` analyzer reports 94% of W3C Canvas 2D API surface used by the bundle is covered shim+bridge. Bug isn't surface-area; it's compositing.
- canvasFillPath dispatches now fire correctly per dispatch log (was 0 before today's gradient-fillRect fix in spectr@def0c9c).
- The bottom toolbar, top toolbar, and bottom strip all render correctly — proves wrap and toolbars composite, but the canvas-children layer doesn't pass parent's bg through.

## What we need

[...]

Issue #972, opened 2026-04-29
view: View::paint_all() doesn't honor z_index() — children paint in insertion order, breaking dropdowns/popovers


danielraffel:
## Repro (Spectr-side)

Spectr's React port at `feature/native-react-editor` HEAD `8f1df47` declares a popover panel like this (`spectr-editor-extracted.js:2847`):

```jsx
<div style={{
  position: 'absolute', top: 28, right: 0, zIndex: 20,
  background: 'rgba(12,16,22,0.96)',
  border: '1px solid rgba(255,255,255,0.12)',
  borderRadius: 4, padding: 3,
  backdropFilter: 'blur(10px)',
}}>...</div>
```

The View receives a `setZIndex(20)` call (per @pulp/css-adapt → bridge), but at paint time the panel renders **behind** sibling content — the `bandsMenu` dropdown is invisible, the PRESETS "MANAGE…" submenu is invisible, etc. Same symptom for every popover in the editor.

## Root cause (Codex spot-check, 2026-04-29)

`core/view/src/view.cpp::View::paint_all()` paints children in **insertion order** and does not sort by `z_index()` before recursing. `setZIndex()` mutates the per-View field but no consumer of that field exists in the paint pipeline.

This is a separate class from #964 (which is about a CanvasWidget surface not reaching the visible compositor) — please do not bundle.

## Acceptance criteria

1. `View::paint_all()` sorts children by `z_index()` ascending before painting (stable sort — siblings with equal z-index keep insertion order).
2. Catch2 regression test in `core/view/test/test_view.cpp` (or wherever paint-order tests live):
   - Parent View with three children A, B, C all at `(0,0,100,100)` with distinct background colors.
   - Set `B.set_z_index(10)` while A and C remain at default `z_index=0`.
   - Render via `RecordingCanvas`; assert the last full-bounds fill is B's color.
   - Stable-order test: two siblings both at `z_index=5` — earlier one paints first.
3. CSS-adapt `zIndex` prop continues to lower to `setZIndex` on the View (already does — no change needed).
4. Screenshot from Spectr's `bandsMenu` test should show the dropdown panel visibly above the canvas after the fix.

## Context

- Umbrella: #924 (Spectr#28 React-bridge port)
- Spectr-side reproduction: `bandsMenu` popover at `spectr-editor-extracted.js:2847` (also any `position:absolute` element with `zIndex: N` in the bundle — multiple sites)
[...]

Issue #969, opened 2026-04-29
view: CSS-style typography inheritance — propagate textColor/fontSize/letterSpacing from parent View to child Label


danielraffel:
Tracked under #924. Spectr#28 native-bridge.

## Problem

Pulp's Label widget honors its own `setTextColor` / `setFontSize` / `setLetterSpacing` calls but doesn't inherit them from a parent View. CSS does: setting `color: white` on a `<div>` cascades to every text descendant unless overridden.

This forces every HTML→Pulp port to either:
- Manually set typography on every leaf Label (tedious, divergent from the source HTML)
- Walk the bundle's React tree at adapt time and push parent typography down (what Spectr does today, in `dom-adapter.tsx`)

## Concrete impact

Spectr's chrome bars set color/fontSize on the toolbar parent and expect children to inherit. The dom-adapter's wrap-strings path explicitly copies `textColor` / `fontSize` / `letterSpacing` to every wrapped Label as a workaround.

## Suggested fix

Two layers:

**Framework (preferred)**:
1. `Label::paint` reads its own `textColor_` / `fontSize_` etc. AND walks up to parent View if unset
2. Parent View stores `inheritable_textColor_` etc. and `setTextColor` on a View propagates the value to the inheritable map (not the View's own paint)
3. Cache resolution per Label so it's a one-time walk on first paint

**Bridge** (minimum):
- `setTextColor(view_id, color)` accepts View IDs (not just Label) — currently silently no-ops on Views
- Add explicit `setInheritedTextColor(id, color)` if framework wants distinct semantics

## Acceptance criteria

- [ ] `<View><Label>x</Label></View>` with `setTextColor(view_id, '#fff')` renders Label as white without explicit `setTextColor(label_id, '#fff')`
- [ ] Same for fontSize, letterSpacing, fontWeight, textAlign
- [ ] Cascade respects override: explicit Label-level value wins over inherited
- [ ] Test `core/view/test/test_typography_inheritance.cpp`
- [ ] Spectr's dom-adapter `inherited*` workaround block (commit 32d4d0f, lines 651–676) can be deleted

danielraffel:
Closed by #1002 (merged 2026-04-29 20:32Z, in v0.63.0). Typography inheritance for Label widgets now framework-level. Spectr-side dom-adapter has manual inheritance code in the wrap-text-children path; that becomes redundant when this lands but doesn't conflict — leaving as-is for back-compat with older SDKs.

PR #973, opened 2026-04-29
test(view): lock in transparent View+CanvasWidget contract (#964 / #967)

danielraffel:
Two regression tests in test/test_canvas_widget.cpp:

1. [issue-964] FilterBank repro — parent View with green setBackground +
   two empty CanvasWidget children. Asserts exactly one full-bounds
   fill_rect is recorded, with the parent's green colour. No opaque
   white, no extra full-bounds fill from any child.

2. [issue-967] Bare View with no setBackground emits zero fill_rect
   commands. HTML <div> default contract.

Both pass against unmodified view.cpp. The "before" snippet in #967's
body (background_color_set_ + kDefaultWhite fallback) does not match
current code — view.cpp:117 already gates the bg fill on has_bg_, which
defaults to false. So no production change is needed; these tests are
contract-only and exist to flag any future regression that reintroduces
an opaque default in View, CanvasWidget, or anywhere along
View::paint_all → CanvasWidget::paint.

The opaque-white symptom #964 reports in Spectr's FilterBank
standalone is therefore not in the View+CanvasWidget paint pipeline.
Diagnosis pivots to the Spectr-side dom-adapter and CanvasWidget
native_gpu_texture_provider path; #964 will be re-titled accordingly.

### DIFF
--- a/test/test_canvas_widget.cpp
+++ b/test/test_canvas_widget.cpp
@@ -504,3 +504,92 @@ TEST_CASE("CanvasWidget draws on top of sibling-painted parent surface",
 }
 
 #endif  // PULP_HAS_SKIA
+
+// pulp #964 / #967 — Spectr's FilterBank scenario, asserted at the
+// command-stream level so the test runs without Skia.
+//
+// A parent View has a green setBackground() applied. It contains two
+// empty CanvasWidget children that occupy the full bounds and have
+// NO draw commands queued — exactly what Spectr's React port produces
+// for FilterBank's two canvas children on first frame.
+//
+// Contract: View::paint_all + CanvasWidget::paint together must NOT
+// emit any opaque-white fill_rect that covers the wrap's bounds. The
+// only fill_rect we expect at full bounds is the parent's green bg
+// fill. If a future change re-introduces an opaque default in View,
+// CanvasWidget, or anywhere between, the recorded stream will pick it
+// up and this test fails.
+TEST_CASE("FilterBank repro: empty canvas children don't paint opaque white",
+          "[canvas_widget][issue-964][issue-967]") {
+    RecordingCanvas rc;
+
+    View wrap;
+    wrap.set_bounds({0, 0, 64, 64});
+    wrap.set_background_color(pulp::canvas::Color::rgba8(0, 200, 0, 255));
+
+    auto cw1 = std::make_unique<CanvasWidget>();
+    cw1->set_bounds({0, 0, 64, 64});
+    auto cw2 = std::make_unique<CanvasWidget>();
+    cw2->set_bounds({0, 0, 64, 64});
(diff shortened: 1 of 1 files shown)


danielraffel:
Spectr-side empirical update before you pivot — the dom-adapter is exonerated:

I dropped an opaque red `fillRect` probe at the very top of FilterBank's `renderAll()` (the canvas2D entry point at `spectr-editor-extracted.js:738`):

```js
ctx.save();
ctx.setTransform(1, 0, 0, 1, 0, 0);
ctx.globalAlpha = 1;
ctx.globalCompositeOperation = 'source-over';
ctx.fillStyle = '#ff0000';
ctx.fillRect(0, 0, w || 1000, h || 1000);
ctx.restore();
```

Standalone re-link, screencap → **red is invisible**. Stdout confirms `canvasFillRect`, `canvasSetTransform`, `canvasSetFillColor` etc. all reach the bridge for `pr_1` (the FilterBank canvas widget) every frame. So the bug is downstream of:

- ✅ JS state (transforms / alpha / composite-op all explicit)
- ✅ dom-adapter (commands reach the bridge correctly)
- ❌ Either CanvasWidget's command replay, native_gpu_texture_provider path, or the surface composition into the parent View's layer

Three audit candidates I'd narrow to:

1. **CanvasWidget surface never flushed** — replay happens into a Skia surface that's not attached to the on-screen layer. Audit: `core/view/src/canvas_widget.cpp::paint()` for any `flush()` / SkSurface lifecycle.
2. **`native_gpu_texture_provider_` path takes over** when it shouldn't — line 29-44 of canvas_widget.cpp branches on `native_gpu_texture_provider_` being set; if `frame.available == true` returns a stale/zero texture and `last_native_gpu_texture_draw_succeeded_` is set, do we still fall through to the command replay loop? Reading the current code, **yes** — but worth confirming the branch returns `false` for "no provider" cases.
3. **`commands_` empty** when expected — bridge dispatch may be writing to a different CanvasWidget instance than the one being painted (id collision / lookup miss).

Posted full repro on #964 ([comment 4341182887](https://github.com/danielraffel/pulp/issues/964#issuecomment-4341182887)).

Separately filed **#972** (filed today) for paint-order/z-index — distinct framework class, also surfaced via this debugging pass. Empirical follow-up there too: absolute children get clipped to parent content bounds (CSS `overflow:visible` not honored), so `#972` may need scope expansion beyond just paint-order sorting. See [#972 comment 4341323470](https://github.com/danielraffel/pulp/issues/972#issuecomment-4341323470).

When this PR merges I'll re-screenshot Spectr against v0.60.x+1 (cron loop is already polling every 30 min for any of #964-#969 + #972 to close).

Approve. Tests are exactly the contract pin we want; no production change required for this PR.

Issue #1009, opened 2026-04-30
release-pipeline: hard-fail fix/feat merges without version bump (codify the #1008 gap)


danielraffel:
## Repro (happened today)

#1008 merged on 2026-04-30 02:36Z — `fix(view): on(id,'click',fn) auto-wires View::on_click (pulp #1006)`. User-facing fix that closes a tracked sub-issue under umbrella #924.

Two hours later — **the fix is still unreleased**. No tag, no SDK tarball, no consumer can pick it up. Auto-release.yml ran at 02:36:48Z and decided not to tag:

```
SDK_AFTER: 0.65.0
SDK_SHOULD_TAG: 0
PLUGIN_AFTER: 0.24.0
PLUGIN_SHOULD_TAG: 0
```

Because no `chore: bump versions` co-commit was created when #1008 merged. The merge went through some path that bypassed `shipyard pr`'s version-bump gate (likely `gh pr merge` direct, or admin-merge through the GitHub UI).

This isn't the first time. Today's merge cadence shows:
- #1004 (RangeSlider) → v0.64.0 ✅ bumped
- #1003 (canvasRect fix) → v0.65.0 ✅ bumped
- #1002 (typography) → v0.63.0 ✅ bumped
- **#1008 (click-dispatch fix) → no bump, no tag** ❌

Spectr (consumer-zero for the React-bridge port) is now stuck on v0.65.0 with the click-dispatch fix sitting unreachable on main.

## Why the existing safety nets didn't catch it

CLAUDE.md "Release Watchdog" describes three layers:

1. `workflow-lint.yml` — PR-gate YAML/structural lint. Doesn't apply (not a workflow change).
2. `auto-release-watchdog.yml` — fires on `auto-release.yml` **failure**. The 02:36 run **succeeded** with `SHOULD_TAG=0`, so the watchdog stayed silent.
3. `release-cadence-check.yml` — every-30-min invariant scanner. Looks for `CMakeLists.txt` VERSION bumps without a matching tag. **Wrong direction** — it catches tag-after-bump gaps but not bump-missing-after-fix gaps.

The hole: a `fix(...)` or `feat(...)` PR can merge without a version bump and every layer reports green.

### Layer 1 — `version-skill-check.yml` hard-fails on fix/feat without bump

Today (per CLAUDE.md): runs in `--mode=report` with `PULP_ENFORCE_PREPUSH=1` in CI. Detects DRIFT, not absence of bump on user-facing PRs.

[...]

PR #1010, opened 2026-04-30
chore: bump versions to v0.66.0 (catches up after #1008)

danielraffel:
## Why

#1008 (`fix(view): on(id,'click',fn) auto-wires View::on_click` — closes pulp#1006) merged on 2026-04-30 02:36Z **without a `chore: bump versions` co-commit**. Auto-release.yml correctly saw no version change and decided not to tag, so the fix is on `main` but unreleased — no tag, no SDK tarball.

Auto-release run: https://github.com/danielraffel/pulp/actions/runs/25144447906 (logs show `SDK_AFTER: 0.65.0`, `SDK_SHOULD_TAG: 0`)

## What this PR does

Bumps SDK 0.65.0 → 0.66.0 so auto-release fires and tags v0.66.0, which triggers release-cli.yml to publish pulp-sdk-darwin-arm64.tar.gz.

## Why this is needed

Spectr (consumer-zero for the React-bridge port) is blocked on validating clicks end-to-end because the SDK still publishes as v0.65.0. The click-dispatch fix is exactly what we need to flip dropdowns / SETTINGS / PRESETS from "render-correct, interaction-dead" to "fully interactive."

## Structural fix

Filed #1009 to harden the release pipeline so this never happens silently again — `version-skill-check.yml` should hard-fail user-facing PRs that lack a version bump, and branch protection should require it. This PR is the manual catch-up; #1009 closes the structural gap.

## Test plan

- Auto-release.yml fires on merge → tags v0.66.0
- release-cli.yml builds and publishes pulp-sdk-darwin-arm64.tar.gz
- Spectr bumps SDK pin to v0.66.0 and validates click dispatch end-to-end

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.65.0
+    VERSION 0.66.0
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C


danielraffel:
Superseded by #1012 — combines this catch-up bump with the canvasFillRect fix for #964 (root cause was a JS-side bridge name mismatch — `web-compat-canvas.js` calls `canvasFillRect` but only `canvasRect` was registered, so every `ctx.fillRect()` silently no-op'd). Single v0.66.0 release will publish both #1008 (click dispatch) and #964 (canvas fillRect).

Issue #640, opened 2026-04-22
coverage: start audio/platform compliance tranche 1


danielraffel:
## Parent
- #641

## Problem

Several audio-critical components remain far below the current tier target.

## Baseline

Captured April 22, 2026 from the public Codecov API:

- audio: 40.97% vs 80 target
- platform: 56.10% vs 80 target
- midi: 65.55% vs 80 target
- signal: 76.33% vs 80 target

Top misses in the first tranche:

- `core/audio/platform/win/wasapi_device.cpp` — 388 misses
- `core/audio/platform/linux/alsa_device.cpp` — 191 misses
- `core/platform/platform/win/child_process_win.cpp` — 154 misses
- `core/midi/platform/mac/coremidi_device.mm` — 184 misses
- `core/midi/platform/win/winmidi_device.cpp` — 169 misses

## Scope

This tranche should focus on deterministic units and existing adjacent tests first, not hardware-bound or host-integration-heavy flows.

Likely starting points:

- `test/test_audio_focus.cpp`
- `test/test_environment.cpp`
- narrowly-scoped platform / MIDI helpers that can be exercised without real devices

## Non-goals

- real device I/O coverage
- broad host/format work already tracked elsewhere (#493)

## Acceptance

- [ ] first tranche lands without introducing hardware-dependent flakes
- [ ] at least one audio-critical component moves materially toward target
- [ ] follow-up slices are narrower and better evidenced than the current umbrella gap

danielraffel:
Phase 2 rebaseline from #723 / Codecov `abe2b07a`:

- `audio`: `20.95%` vs `80%` target, gap `59.05`
- `platform`: `35.41%` vs `80%` target, gap `44.59`

Largest represented misses to rank for tranche 1:
- `core/audio/platform/win/wasapi_device.cpp`: 486 misses
- `core/audio/platform/linux/alsa_device.cpp`: 251 misses
- `core/audio/src/aiff_reader.cpp`: 199 misses
- `core/audio/src/format_registry.cpp`: 161 misses
- `core/audio/platform/mac/coreaudio_device.mm`: 120 misses
- `core/platform/platform/win/child_process_win.cpp`: 192 misses
- `core/platform/platform/mac/file_dialog_mac.mm`: 84 misses
- `core/platform/platform/linux/clipboard_linux.cpp`: 55 misses

[...]

PR #1021, opened 2026-04-30
test(platform): cover child process edge paths

danielraffel:
Parent tracker: #640
Umbrella tracker: #641

## Scope
- Add child-process edge coverage for partial stdout without a trailing newline, repeated wait() after completion, and timeout output preservation.
- Codify the Phase 3 no-idle coverage cadence in docs/guides/coverage.md: monitor PRs, debug failures immediately, and keep advancing focused tranches while CI-only blockers run.

## Validation
- cmake --build build --target pulp-test-child-process -j$(sysctl -n hw.ncpu)
- ./build/test/pulp-test-child-process
- ctest --test-dir build -R "child_process|ChildProcess|exec captures|timeout|wait|line callback|stderr|find_on_path|working directory|output capture" --output-on-failure
- git diff --check

## CI
- Opened with shipyard pr, deliberately skipping local VM targets so Namespace/GitHub checks are the evidence path.
- Codecov patch check expected on this PR.

### DIFF
--- a/docs/guides/coverage.md
+++ b/docs/guides/coverage.md
@@ -220,6 +220,19 @@ at a **75%** floor. Sub-threshold diff coverage hard-fails this check
 and blocks the merge — adding untested code means either adding tests
 or splitting the untested portion into its own PR.
 
+### Phase 3 PR cadence
+
+Phase 3 is tranche work, not a queue stall. While a Codecov remediation
+PR is pending and blocked only on CI, keep moving with another small,
+focused coverage tranche. Check the open PR list and GitHub checks
+before starting each tranche, watch them again after pushing it, debug
+new failures as soon as they appear, and manually merge PRs once they
+are green.
+
+Use Namespace-backed GitHub checks and Codecov comments as the default
+evidence path. Local VMs are a fallback for reproducing or isolating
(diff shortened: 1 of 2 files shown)


danielraffel:
Pushed `69f1a08f` to address the Windows Namespace failure. Root cause was the Windows-only partial-stdout fixture command (`<nul set /p dummy=partial`) producing the intended no-newline stdout but exiting with code 1 on the runner. The fix keeps the same no-newline stdout behavior and explicitly exits 0. Local `git diff --check` passed; the fresh Windows Namespace rerun is the relevant validation for this path.

danielraffel:
Follow-up pushed: 30a5cc19. The rerun still failed Windows Namespace on `line callback buffers partial stdout without trailing newline`; the log showed stdout was `partial `, not `partial`. Root cause was the space before `& exit /b 0` being included in cmd.exe's `set /p` prompt. The fixture now uses `set /p dummy=partial<nul& exit /b 0` to emit no trailing newline and no prompt-space while preserving exit 0.

Local build was not rerun in the fix worktree because its build dir is not configured; the hosted Windows Namespace rerun is the target validation.

Issue #1049, opened 2026-04-30
[epic] Audit silent-no-op config-driven surfaces (post-#1005)


danielraffel:
# [epic] Audit silent-no-op config-driven surfaces (post-#1005)

**Status:** open until all child issues close.

## The bug class

A config option exists, the script that reads it builds CLI args from it, but those args are silently ineffective. No test catches it because the option is rarely-touched. Adding a second entry / changing context exposes the bug.

The canonical example is **`diff_cover_excludes`** in `tools/scripts/coverage_config.json` — surfaced and fixed on PR #1005:

- **The bug:** `--exclude=foo --exclude=bar` is `nargs='+'` argparse → keeps only the LAST. AND `core/host/src/scanner_clap.cpp` (literal relative path) matches NEITHER basename nor abspath via fnmatch → silent no-op. The exclude config has been ineffective since #919; nobody noticed because there was only one entry and it didn't need to do anything.
- **The fix** (commit `efefe144`): pass all entries under a single `--exclude val1 val2 ...` flag; convert config entries to `**/<basename>` glob form; document the contract.
- **The lock-in** (commit `b258730c`): three structural tests in `DiffCoverExcludeContractTests`. Reverting any part of the fix makes the relevant test fail loudly with an actionable message.

This issue tracks the audit of **other config-driven surfaces** that may have the same shape risk.

## Children

Each child issue audits ONE surface. Self-contained — an agent reading any child has everything needed to audit, fix, test, PR, and close.

- [ ] #1052 — audit: `coverage_config.json` filters / exclude_filters for silent-no-op
- [ ] #1053 — audit: `skill_path_map.json` no-match case + glob format
- [ ] #1054 — audit: `versioning.json` bypass-trailer parsing edge cases
- [ ] #1055 — audit: `codecov.yml` component classification — every surface buckets exactly once
- [ ] #1056 — audit: `ci/coverage-targets.yaml` tier coverage gaps
- [ ] #1057 — audit: skill `requires:` arrays — runtime presence of declared deps
- [ ] #1058 — audit + fix: thread `lcov --remove` through coverage pipeline (LCOV_EXCL propagation)
- [ ] #1144 — audit + fix: flip pre-push gates from advisory to enforcing by default (the churn-meta — would have collapsed PR #1005's 7 iterations to 1)

## Rules of engagement (for any agent picking up a child)

**You SHOULD:**
1. Read the child issue's audit recipe and run it
[...]

PR #1005, opened 2026-04-29
feat(rust-cli): Phase 8 integrated prep — coverage uplift, parity, scan hardening, dual-binary tarball

danielraffel:
## DO NOT MERGE WITHOUT EXPLICIT GO-AHEAD

Phase 8 cutover protocol: this PR updates and validates the side-by-side Rust CLI migration work, but it does **not** flip the default `pulp` binary. Merge is still a human call.

## Current State

- Branch: `validate/phase8-integrated`
- Final merge-prep rebase onto `origin/main` at `f58b18e0` on 2026-05-05
- Head pushed: `8a72fb51`
- This PR keeps C++ as the default `pulp` and ships Rust as `pulp-rs`
- The default flip is **not** old PR #783. #783 is stale and should not be used.
- The prepared flip branch is `feature/phase8-default-flip-prepped`; after #1005 merges it should be rebased and opened as a new draft flip PR.

## What This PR Contains

- Rust CLI Phase 8 prep and parity work for the `validate/phase8-integrated` branch
- Rust absorption of recent C++ CLI features including `run --headless/--screenshot/--frames/--watch`
- Rust-owned `upgrade --install` path and dual-binary packaging prep
- C++ scan hardening and `scan --no-load`
- Cross-binary config/projects/project-bump parity
- Sandbox E2E coverage for side-by-side swap, rollback, cross-binary state, and fallthrough
- Drift patch for current `main`: Rust help includes `loop`, `coverage`, and `harness`; `inspect` summary/order matches current C++
- `pulp harness` remains C++-owned and is bridged by Rust fallthrough until/unless deliberately ported

## Harness Coordination

Harness work through #1395, #1397, #1398, #1399, #1415, and #1422 is now on `main`. This PR was rebased after those commits.

#1005 keeps runtime ownership for `pulp harness` on the C++ side and only mirrors the user-visible command surface in Rust:

- `experimental/pulp-rs/src/help.rs` lists `harness` in the script-command help table
- the Rust help fixture includes the C++ `harness` help entry
- sandbox fallthrough coverage proves `pulp harness --help` reaches `pulp-cpp`
- `docs/status/cli-commands.yaml`, `docs/reference/cli.md`, and the CLI sync skip list include `harness`
- local harness smoke now sees all auto-discovered surfaces: `canvas2d`, `css`, `html`, `yoga`

Keep the Phase 8 dual-binary packaging surfaces coordinated:

- `tools/scripts/package_cli.py`
- `.github/workflows/release-cli.yml`
[...]

Issue #290, opened 2026-04-16
test-coverage hardening: raise low-ratio subsystems + audit placeholder tests


danielraffel:
From the 2026-04-16 alpha-hardening audit:

- Static test/code LOC ratio: 0.261 across 744 code files / 151,726 LOC.
- `core/dsl`: **0.00** — header-only subsystem with no tests. Confirmed: only `include/` + `CMakeLists.txt`.
- Weak-surface subsystems (priority order): `core/platform` 0.28, `core/host` 0.32, `core/osc` 0.37, `core/runtime` 0.44, `core/view` 0.51, `core/format` 0.52.

## Scope (prioritized, tackle toward end of production-readiness push)

1. **Audit placeholder tests** flagged as succeed-only scaffolding. Decide per file: make real or delete.
   - `test_validation_harness.cpp`, `test_text_input.cpp`, `test_uia_mapping.cpp`,
     `test_negative_path.cpp`, `test_tempo_hook.cpp`, `test_showcase.cpp`,
     `test_js_engine.cpp`, `test_websocket_channel.cpp`, `test_atk_mapping.cpp`,
     `test_announce.cpp`, `test_memory_pressure.cpp`, `test_transport_hook.cpp`,
     `test_osc_channel.cpp`, `test_widget_bridge.cpp`, `test_audio_system_hotplug.cpp`,
     `test_network_stream.cpp`, `test_screenshot.cpp`, `test_token_diff.cpp`.
2. **Priority subsystem hardening** — raise coverage on: **platform, host, format, runtime, view, osc, dsl** (user-specified priority). Each gets a focused sweep: identify the 3-5 highest-blast-radius entry points, write behaviour tests.
3. **core/dsl**: decide whether it's live. If yes, add parser + interpreter suite. If dead, delete the subsystem.
4. **Regression-test policy**: every new bug-fix PR carries a narrow regression test or an explicit "no test practical" note.
5. **Runtime coverage job**: add an LLVM/gcov CI lane at a later phase; don't block current bugs on coverage infra.

Schedule: tackle after production-readiness workstream PRs have drained (per user: "obv towards end").

danielraffel:
Placeholder-test audit walked all 18 flagged files in test/:

**Real placeholders** (should be filled or removed):
- `test_showcase.cpp` (6 lines, just `REQUIRE(true)`) — comment says "will be populated when design tool merges". Still pending.
[...]

PR #1152, opened 2026-04-30
fix(coverage): remove unread filters/exclude_filters keys from coverage_config.json

### LINKED ISSUE #1052: audit: coverage_config.json filters/exclude_filters for silent-no-op
# audit: coverage_config.json `filters` / `exclude_filters` for silent-no-op

Parent: #1049

## Surface

`tools/scripts/coverage_config.json`:

```json
"filters": ["core/**", "tools/cli/**", "tools/scripts/**"],
"exclude_filters": ["external/**", "build/**", "test/**", "examples/**", "experimental/**"]
```

These are claimed to control which files diff-cover analyzes. Same family as `diff_cover_excludes` (which was silently broken — see PR #1005 / commit `efefe144`).

## Audit recipe

```bash
cd /Users/danielraffel/Code/pulp
grep -nE 'filters|exclude_filters' tools/scripts/local_diff_cover.sh .github/workflows/coverage.yml
```

Look for: where these arrays are read, how they're translated into CLI args (or other behavior), and whether they hit the same `--exclude=foo --exclude=bar` (overwrite) or "literal-path-vs-fnmatch" issues.

Then run the cobertura coverage flow against a known fixture and verify whether files in `exclude_filters` actually get filtered:

```bash
xcrun --find llvm-cov   # ensure llvm tools on PATH
PULP_SKIP_RUST_CLI=1 ~/.pulp/bin/pulp coverage diff
# Inspect /Users/danielraffel/Code/pulp/build-cov/coverage.cobertura.xml — do `external/**` files appear?
grep 'filename="external/' /Users/danielraffel/Code/pulp/build-cov/coverage.cobertura.xml | head -3
```

If `external/**` files show up in the cobertura XML despite being in `exclude_filters`, the filter is not being applied.

## Bug-detection criteria

**BAD** (silent no-op):
- `exclude_filters` is read but never passed to anything that filters
- `exclude_filters` is read and passed, but pattern format doesn't match (literal-vs-glob issue)
- Repeated arg shape that argparse overwrites

**GOOD**:
- `exclude_filters` actively reduces the cobertura XML's filename set
- A test fixture proves it (e.g., `test_local_diff_cover.py::FiltersAreApplied`)

## Fix template

If the bug exists, mirror **commit `efefe144`** (`fix(coverage): diff_cover_excludes was a silent no-op for any entry with a slash`):

- Fix the pattern format if it's literal-vs-fnmatch
- Fix the arg shape if it's argparse-overwrite
- Document the contract inline in `coverage_config.json`'s `_comment`
- Update both `tools/scripts/local_diff_cover.sh` AND `.github/workflows/coverage.yml`

## Test contract (mandatory per #290)

Mirror **commit `b258730c`** — add a class to `tools/scripts/test_local_diff_cover.py`:

```python
class FilterContractTests(unittest.TestCase):
    def test_exclude_filters_actually_filter_cobertura_xml(self):
        # Fixture: a synthetic cobertura with external/** entries.
        # Run the local script's pre-diff-cover XML generation.
        # Assert no external/** filename remains in the output XML.
```

Anti-drift required — a failing fixture must fail the test loudly with a message pointing at this issue.

## Acceptance

- [ ] Audit recipe run; output captured in a comment on this issue
- [ ] If clean: close with audit evidence
- [ ] If buggy: PR opened with fix + test
- [ ] PR body contains `Closes #<this issue>` and `Refs #1049`
- [ ] Test verified to fail when the fix is reverted (see `b258730c` for the regression-check pattern)

## Boundaries

- **DO NOT touch:** `diff_cover_excludes` (already fixed), Codecov component definitions, tier definitions, version-bump checks
- **DO NOT** lower `diff_coverage_fail_under`
- **DO NOT** add files to `diff_cover_excludes` to mask coverage gaps in `external/**` etc.

## Merge policy

- Open PR via `gh pr create`
- DO NOT run `gh pr merge` — humans own the merge step

## Refs

- Parent: #1049
- Pattern: PR #1005, commits `efefe144` + `b258730c`

danielraffel:
## Summary

`filters` and `exclude_filters` lived in `tools/scripts/coverage_config.json` but were never read by anything — neither `tools/scripts/local_diff_cover.sh` nor `.github/workflows/coverage.yml` referenced them. A user who edited `exclude_filters` to skip `external/**` saw zero effect — silent no-op.

Same family of bug as the `diff_cover_excludes` issue fixed in `efefe144` / locked-in by `b258730c` (PR #1005), but in this case the entries weren't even reaching diff-cover — there was simply no consumer at all. The implied "these arrays drive what diff-cover analyzes" contract didn't exist.

Source-set filtering for `llvm-cov` already lives in `COVERAGE_IGNORE_REGEX` inside `local_diff_cover.sh` and `scripts/run_coverage.sh`, so removing the JSON arrays does not lose any real behavior — it only removes documentation that lied about a contract.

## Changes

- Drop `filters` and `exclude_filters` from `coverage_config.json`.
- Update `_comment` to enumerate the actually-consumed keys (`diff_coverage_fail_under`, `compare_branch`, `diff_cover_excludes`) and call out that source-set filtering belongs in `COVERAGE_IGNORE_REGEX`, not in this JSON.

## Test contract (mandatory per #290)

New `ConfigKeysAreConsumed` class in `tools/scripts/test_local_diff_cover.py`:

- `test_every_config_key_has_a_consumer` — every non-underscore key in the JSON must appear by name in `local_diff_cover.sh` or `coverage.yml`. Underscore-prefixed keys (`_comment`, `_meta`, ...) are exempted as documentation-only metadata. Future dead config additions fail loudly with a pointer to #1052.
- `test_silent_noop_filter_keys_stay_removed` — belt-and-suspenders explicit reject for the two specific keys that triggered this audit, with an actionable failure message naming the issue.

## Anti-drift verification

Re-added `filters: [core/**]` + `exclude_filters: [external/**]` to the JSON locally — both tests failed loudly with messages pointing at #1052 (and an actionable remediation hint: wire it up, remove it, or rename with `_` prefix). Removing the keys (the fix) makes them pass.

## Boundaries

- Did NOT touch `diff_cover_excludes` (already fixed by #1005)
- Did NOT touch Codecov component definitions, tier definitions, or version-bump checks
- Did NOT lower `diff_coverage_fail_under`
- Did NOT add files to `diff_cover_excludes` to mask any coverage gap

## Test plan

- [x] `python3 tools/scripts/test_local_diff_cover.py` — all 10 tests pass with the fix applied
- [x] Reverted the fix locally (re-added the two keys) — both new tests fail loudly with #1052 in the message
- [x] Restored the fix — all tests green again

Closes #1052
Refs #1049


danielraffel:
## Codex review (second opinion)

- `test_every_config_key_has_a_consumer` is weaker than its name: `if key in script_text or key in workflow_text` passes on comments, docs, or error strings, not actual reads. Use stricter patterns like `read_config_value <key>` and `jq -r '.<key>'`.

- The removal leaves stale downstream/user-facing consumers saying "threshold + filters" still live in `coverage_config.json`: `tools/cli/cmd_coverage.cpp:6`, `CLAUDE.md:611`, `.agents/skills/ci/SKILL.md:794`, `.agents/skills/cli-maintenance/SKILL.md:345`. That directly affects downstream-project skills and could cause the dead keys to be reintroduced.

_Generated via `codex exec` (low reasoning effort, 5-min timeout). Review prompt asked for max 5 bullets; only real issues flagged._

PR #1156, opened 2026-04-30
fix(coverage): honor LCOV_EXCL markers via lcov_strip_excl.py (#1058)

### LINKED ISSUE #1058: audit + fix: thread lcov --remove through coverage pipeline (LCOV_EXCL propagation)
# audit + fix: thread `lcov --remove` through coverage pipeline so LCOV_EXCL markers reach diff-cover

Parent: #1049

## Surface

`scripts/run_coverage.sh` and `tools/scripts/local_diff_cover.sh` generate cobertura XML via `llvm-cov export --format=lcov` → `lcov_cobertura.py`. The `LCOV_EXCL_*` markers in C++ source code are gcov-specific; **`llvm-cov-export` doesn't honor them**. Result: code wrapped in `LCOV_EXCL_START..STOP` STILL appears in the diff-cover report as missing — making the markers documentation-only.

This was discovered on PR #1005 (commit `efefe144`'s context). The current workaround is `**/scanner_clap.cpp` in `diff_cover_excludes` — file-level exclusion that's coarser than the LCOV markers want.

## Audit recipe (verifies the bug exists)

```bash
cd /Users/danielraffel/Code/pulp
grep -nE 'LCOV_EXCL' core/host/src/scanner_clap.cpp | head -5
xcrun --find llvm-cov
PULP_SKIP_RUST_CLI=1 ~/.pulp/bin/pulp coverage diff 2>&1 | grep -A 2 'scanner_clap'
# Look at the missing-lines list — if line numbers inside an LCOV_EXCL_START..STOP
# block appear, the markers aren't propagating.
```

If lines marked `LCOV_EXCL_START..STOP` show up as "Missing", the toolchain bug is confirmed.

## Bug-detection criteria

**BAD** (current state): line numbers inside LCOV_EXCL blocks appear in diff-cover output.
**GOOD**: lines inside LCOV_EXCL blocks are filtered out before diff-cover sees them.

## Fix template

This one is a **toolchain fix**, not a config fix. Two paths:

### Path A — Insert `lcov --remove` between llvm-cov export and Cobertura conversion

```bash
# Current shape (in local_diff_cover.sh and scripts/run_coverage.sh):
llvm-cov export --format=lcov ... > coverage.lcov
python3 lcov_cobertura.py coverage.lcov --output coverage.cobertura.xml

# Proposed shape:
llvm-cov export --format=lcov ... > coverage.raw.lcov
lcov --remove coverage.raw.lcov '*' --output-file coverage.lcov  # honors LCOV_EXCL
python3 lcov_cobertura.py coverage.lcov --output coverage.cobertura.xml
```

`lcov` is part of the LCOV package (separate from `llvm-cov`). Available via Homebrew (`brew install lcov`) on macOS, `apt install lcov` on Linux.

### Path B — Use a python helper to strip LCOV_EXCL'd lines from the .lcov file

Avoids the lcov dependency. ~50 lines of Python that walk the .lcov, identify SF/DA records inside an LCOV_EXCL range, and elide them.

Pick whichever lands cleaner on CI.

## Test contract (mandatory per #290)

Add to `tools/scripts/test_local_diff_cover.py` or a new `test_lcov_propagation.py`:

```python
class LcovExclPropagation(unittest.TestCase):
    def test_lines_inside_lcov_excl_block_do_not_appear_in_cobertura(self):
        # Fixture: a small .cpp + corresponding lcov-info that mimics
        # what llvm-cov-export produces, with one block bracketed by
        # LCOV_EXCL_START..STOP. After running the pipeline, those
        # lines must not appear in the resulting cobertura XML.
```

## Acceptance

- [ ] Audit recipe run; bug confirmed (or refuted) with output in a comment
- [ ] If confirmed: PR landing the chosen fix path + structural test
- [ ] After merge: drop `**/scanner_clap.cpp` from `coverage_config.json`'s `diff_cover_excludes` (no longer needed)
- [ ] Test verified to fail when the propagation step is removed

## Boundaries

- **DO NOT** drop scanner_clap.cpp from `diff_cover_excludes` until the propagation fix is verified working in CI
- **DO NOT** change the diff-cover threshold
- **DO NOT** add lcov as a hard dep on developer machines without the lazy-prompt pattern (mirror `local_diff_cover.sh`'s existing dep-preflight)

## Merge policy

- Open PR via `gh pr create`
- DO NOT run `gh pr merge`

## Refs

- Parent: #1049
- Pattern: PR #1005, commits `efefe144` + `b258730c`
- Documented in `tools/scripts/coverage_config.json` `_comment` as a known toolchain gap

danielraffel:
## Summary

`llvm-cov export --format=lcov` does NOT honor `LCOV_EXCL_LINE` / `LCOV_EXCL_START..STOP` markers — those are a `gcov`/`lcov` convention, not an LLVM one. Without a strip pass between `llvm-cov export` and the Cobertura converter, code wrapped in `LCOV_EXCL` still appears in diff-cover's missing-lines report and in the Cobertura XML uploaded to Codecov, making the markers documentation-only.

This was the gap that forced PR #1005's `**/scanner_clap.cpp` file-level workaround in `diff_cover_excludes` (since reverted on main), and it's the toolchain bug tracked in #1058.

## Audit confirmation

- `grep LCOV_EXCL core/host/src/scanner_clap.cpp` on `origin/main` → 0 hits (the LCOV_EXCL bracketing from PR #1005 was rolled back when the file-level exclude was rolled back). The bug itself is structural: `llvm-cov-export` ignores LCOV markers regardless of where they live, so any future LCOV_EXCL anywhere in the tree would silently fail to propagate.
- `**/scanner_clap.cpp` is **not** currently in `tools/scripts/coverage_config.json`'s `diff_cover_excludes` on `origin/main`, so there is nothing to remove in this PR (per the issue's "DO NOT drop" boundary, which is already satisfied).
- Verified by reverting the strip wire in `local_diff_cover.sh` and running the new wiring test: `AssertionError: -1 not greater than 0 : no `python3 ... lcov_strip_excl.py` invocation`. Restoring the wire makes the test pass.

## Path chosen — Path B (Python helper)

Picked over Path A (insert `lcov --remove`) because:

- The issue's **Boundaries** explicitly call out: "DO NOT add lcov as a hard dep on developer machines without the lazy-prompt pattern." Path B avoids the dep entirely.
- Cross-platform: `lcov` is non-trivial on Windows, easy on macOS via Homebrew, easy on Linux via apt — Python is already on every developer machine and every Pulp CI lane.
- Pulp already vendors `tools/scripts/lcov_cobertura.py` (a Python script), so this fits the existing pattern.
- Pure-Python is faster to test (we ship 16 fixture tests that run in ~0.1s each).

The cost is ~330 lines of Python to maintain. Acceptable given the alternative is a multi-platform dep-preflight + brew/apt install paths in three scripts.

## Changes

- **`tools/scripts/lcov_strip_excl.py`** (new) — pure-Python helper. Walks each `SF:<path>` record, opens the source, scans for `LCOV_EXCL_*` markers, and drops `DA:` / `BRDA:` / `FN[A]:` / `FNDA:` records on excluded lines. Recomputes `LF` / `LH` / `BRF` / `BRH` / `FNF` / `FNH` counters. Honors `LINE`, `START`, `STOP`, `BR_LINE`, `BR_START`, `BR_STOP`. Orphan markers warn but degrade safely (lcov(1)-compatible behavior).
- **`tools/scripts/local_diff_cover.sh`** — wires the strip step between `llvm-cov export` and the Cobertura converter. Adds an existence preflight on `lcov_strip_excl.py`.
- **`scripts/run_coverage.sh`** — same wiring on the CI lane.
- **`tools/scripts/test_lcov_strip_excl.py`** (new) — 16 fixture tests:
  - Marker-grammar tests (`compute_exclusions` over each marker variant + edge cases).
  - `.lcov` filter tests (DA/BRDA drops, LF/LH recompute, no-marker passthrough, unreadable-source passthrough).
  - **End-to-end test** (the headline acceptance criterion from #1058): synthesize a source with `LCOV_EXCL_START..STOP`, synthesize the matching `.lcov`, run the strip script as a subprocess, run the Cobertura converter, parse the XML, assert excluded line numbers do NOT appear as `<line number="..."/>` elements.
  - Pipeline-wiring tests: both shell scripts must invoke the strip helper, and the strip step must run BEFORE the Cobertura converter.

## Acceptance (from #1058)

- [x] Audit recipe run; bug confirmed (structural — LCOV_EXCL markers don't propagate through `llvm-cov export`).
- [x] PR landing the chosen fix path + structural test.
- [x] Test verified to fail when the propagation step is removed (reverted strip wire → wiring test fails with actionable message).
- [ ] **Follow-up after merge**: the `**/scanner_clap.cpp` entry can be removed from `coverage_config.json`'s `diff_cover_excludes` once it's needed again. It's NOT in the excludes list on `origin/main` today, so there's nothing to remove in this PR. Mention in the changelog when the strip path is verified working in CI.

## Boundaries respected

- Did NOT drop anything from `coverage_config.json`'s `diff_cover_excludes` (the boundary).
- Did NOT change the diff-cover threshold.
- Did NOT add `lcov` as a hard dep — Path B uses pure Python.
- Tests ship with the fix in this PR, per CLAUDE.md "Tests ship with fixes".

## Test plan

- [x] `python3 tools/scripts/test_lcov_strip_excl.py` — 16/16 pass locally.
- [x] `python3 tools/scripts/test_local_diff_cover.py` — 8/8 still pass (no regression).
- [x] `python3 tools/scripts/test_lcov_cobertura.py` — 3/3 still pass.
- [x] `python3 tools/scripts/test_run_coverage.py` — 13/13 still pass.
- [x] CLI sanity check: synthesized .lcov with LCOV_EXCL'd lines → strip script removes them and recomputes LF/LH correctly.
- [ ] CI's coverage workflow (`.github/workflows/coverage.yml`) still produces a non-empty Cobertura XML on the macOS / Linux / Windows lanes.

Closes #1058
Refs #1049

### DIFF
--- a/docs/guides/coverage.md
+++ b/docs/guides/coverage.md
@@ -292,6 +292,17 @@ Source → Gradle `testDebugUnitTest`
   `llvm-cov export --format=lcov` plus the vendored
   `tools/scripts/lcov_cobertura.py` converter emit Cobertura XML for
   Codecov + diff-cover.
+- **`LCOV_EXCL` markers**: `tools/scripts/lcov_strip_excl.py` runs
+  between `llvm-cov export` and `lcov_cobertura.py` (wired into both
+  `tools/scripts/local_diff_cover.sh` and `scripts/run_coverage.sh`).
+  llvm-cov itself does NOT honor `LCOV_EXCL_LINE` /
+  `LCOV_EXCL_START..STOP` (those are a gcov/lcov convention, not LLVM),
+  so without this strip pass the markers are documentation-only —
+  excluded code still appears in diff-cover's "missing lines" report
+  and in the Cobertura XML uploaded to Codecov. The Python helper
+  walks each `SF:<path>` record, drops `DA`/`BRDA`/`FN[A]`/`FNDA` rows
+  on excluded lines, and recomputes the `LF`/`LH`/`BRF`/`BRH`/`FNF`/`FNH`
+  counters so the output is internally consistent. See `#1058`.
 - **`-object` list**: llvm-cov only reports translation units linked
   into the binaries given via `-object`. The script passes every test
   executable **plus** every first-party static archive (`libpulp-*.a`)
--- a/scripts/run_coverage.sh
+++ b/scripts/run_coverage.sh
@@ -318,21 +318,36 @@ llvm-cov show \
 # a vendored copy of https://github.com/eriwen/lcov-to-cobertura-xml
 # (Apache 2.0, MIT-compatible).
 COBERTURA_XML="${BUILD_DIR}/coverage.cobertura.xml"
+LCOV_RAW="${REPORT_DIR}/coverage.raw.lcov"
 LCOV_FILE="${REPORT_DIR}/coverage.lcov"
 LCOV_COBERTURA="${REPO_ROOT}/tools/scripts/lcov_cobertura.py"
+LCOV_STRIP_EXCL="${REPO_ROOT}/tools/scripts/lcov_strip_excl.py"
 
 if [[ ! -f "${LCOV_COBERTURA}" ]]; then
     echo "run_coverage.sh: missing ${LCOV_COBERTURA}" >&2
     echo "    (expected the vendored lcov→cobertura converter)" >&2
     exit 1
 fi
+if [[ ! -f "${LCOV_STRIP_EXCL}" ]]; then
+    echo "run_coverage.sh: missing ${LCOV_STRIP_EXCL}" >&2
+    echo "    (expected the LCOV_EXCL stripper — pulp #1058)" >&2
+    exit 1
+fi
 
-echo "=== llvm-cov export → LCOV ==="
+echo "=== llvm-cov export → LCOV (raw) ==="
 llvm-cov export --format=lcov \
     "${BINARIES[@]}" \
     -instr-profile="${PROFDATA}" \
     -ignore-filename-regex="${COVERAGE_IGNORE_REGEX}" \
-    > "${LCOV_FILE}"
+    > "${LCOV_RAW}"
+
+# llvm-cov-export does NOT honor LCOV_EXCL_LINE / LCOV_EXCL_START..STOP
+# markers — those are a gcov/lcov convention. Without this strip step,
+# code wrapped in LCOV_EXCL still appears in the diff-cover "missing
+# lines" report and in the Cobertura XML uploaded to Codecov
+# (pulp #1058).
+echo "=== Strip LCOV_EXCL ranges → LCOV (filtered) ==="
+python3 "${LCOV_STRIP_EXCL}" "${LCOV_RAW}" "${LCOV_FILE}"
 
 echo "=== LCOV → Cobertura XML ==="
 python3 "${LCOV_COBERTURA}" "${LCOV_FILE}" \
--- a/tools/scripts/lcov_strip_excl.py
+++ b/tools/scripts/lcov_strip_excl.py
@@ -1,0 +1,506 @@
+#!/usr/bin/env python3
+# tools/scripts/lcov_strip_excl.py
+#
+# Strip lines wrapped in `LCOV_EXCL_*` markers from an .lcov file before
+# it reaches the Cobertura converter.
+#
+# Why this script exists
+# ----------------------
+# `llvm-cov export --format=lcov` does NOT honor `LCOV_EXCL_LINE` /
+# `LCOV_EXCL_START` / `LCOV_EXCL_STOP` markers. Those markers are a
+# `gcov` / `lcov` convention; LLVM's exporter just emits the raw line
+# data and the markers are silently ignored. Result: any code wrapped in
+# `LCOV_EXCL_START..STOP` STILL appears in `diff-cover`'s "missing
+# lines" report — making the markers documentation-only.
+#
+# This helper closes that gap. It walks each `SF:<path>` record in the
+# input .lcov, opens the source file, scans for the LCOV_EXCL markers,
+# and drops every `DA:<line>,...`, `BRDA:<line>,...`, and `FN[A]:<line>,...`
+# record whose line falls inside an excluded range. The line/branch/fn
+# totals (`LF`, `LH`, `BRF`, `BRH`, `FNF`, `FNH`) are recomputed so the
+# resulting file is internally consistent.
+#
+# Markers honored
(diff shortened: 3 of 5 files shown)


danielraffel:
## Codex second-opinion review (gpt-5-codex, low reasoning)

- **Issue: source read failures silently disable stripping.** `_read_source()` returns `None` on any `OSError`, and `filter_lcov()` treats that as an empty exclusion map. If `SF:` paths are relative, remapped, deleted, generated, or otherwise not readable from the script's cwd, the strip step becomes a silent pass-through. That directly affects the intended guarantee; at minimum warn per unreadable `SF:` path, or support `--base-dir` / repo-root resolution.

- **Issue: nested `LCOV_EXCL_START` blocks are not handled as nested.** `compute_exclusions()` uses a boolean `in_block`, so a second `LCOV_EXCL_START` inside an active block is effectively forgotten, and the first following `LCOV_EXCL_STOP` ends exclusion for the outer block too. Same issue for `LCOV_EXCL_BR_START`. If nested markers are considered malformed, this should warn; if supported, use a depth counter.

- **Counters are mostly recomputed, not just deleted.** `LF/LH`, `BRF/BRH`, and `FNF/FNH` are regenerated from kept records. One caveat: `FNF` counts both `FN:` and `FNA:` records as found functions, so mixed `FN` + `FNA` input for the same function would overcount. If this pipeline only emits normal `FN/FNDA`, it's fine.

- **Marker grammar covers the requested forms.** The regex recognizes `LCOV_EXCL_LINE`, `LCOV_EXCL_START`, `LCOV_EXCL_STOP`, `LCOV_EXCL_BR_LINE`, `LCOV_EXCL_BR_START`, and `LCOV_EXCL_BR_STOP`, with word-boundary protection against suffix matches.

- **Script wiring looks idempotent in the diff.** Both shell scripts export to `coverage.raw.lcov`, strip into `coverage.lcov`, then convert that filtered file. Running the pipeline twice regenerates raw input first, so it should not repeatedly strip an already-filtered LCOV file.

---
*Generated via `codex exec` (gpt-5-codex, `--reasoning-effort=low`). Review-only — no merge, no push.*

danielraffel:
Closing as duplicate — main shipped equivalent fix via #1326 (the upstream issue is closed). Verified: same target issue, same files modified, comparable scope. If a piece of this PR was actually distinct from #1326, it can be re-filed against current main.

PR #1168, opened 2026-04-30
fix(pulp-react): auto-register hover dispatch when onMouseEnter is set (#1149)

### LINKED ISSUE #1149: @pulp/react: hover + tap-state parity — registerHover not called from prop-applier; no CSS :hover translation; e.currentTarget.style mutations no-op
## Summary

WebView builds get hover and tap-state styling for free (CSS `:hover`, `:active`, `e.currentTarget.style.X` mutations, React state-based flash patterns). Native @pulp/react builds get **none** of this end-to-end on v0.68.0:

- CSS `:hover` rules → not parsed, not translated.
- `onMouseEnter` / `onMouseLeave` React handlers → registered as JS listeners, but **the bridge never fires the events** because `registerHover(id)` is never called.
- `e.currentTarget.style.X = Y` imperative pattern → no DOM, silently no-ops.
- React state-based tap flash (`[flash, setFlash]` + `setTimeout`) → works *if* the click dispatches (so it works for top-level toolbar buttons post-#1073, but not inside popovers — that's blocked on #1148).

This is the third distinct "interaction parity" framework gap on the Spectr port — sister to #1147 (popover render) and #1148 (popover click dispatch).

### `WidgetBridge` already supports hover dispatch — opt-in
`core/view/src/widget_bridge.cpp:1356-1370`:

```cpp
// registerHover(id) — enables "mouseenter"/"mouseleave" JS callbacks (CSS :hover)
engine_.register_function("registerHover", [this](choc::javascript::ArgumentList args) {
    auto id = args.get<std::string>(0, "");
    auto it = widgets_.find(id);
    if (it != widgets_.end()) {
        auto alive = callback_alive_;
        auto* engine = &engine_;
        it->second->on_hover_enter = [alive, engine, id]() {
            safe_dispatch_eval(alive, engine, "__dispatch__('" + id + "', 'mouseenter', 0)", "hover enter");
        };
        it->second->on_hover_leave = [alive, engine, id]() {
            safe_dispatch_eval(alive, engine, "__dispatch__('" + id + "', 'mouseleave', 0)", "hover leave");
        };
    }
    return choc::value::Value();
});
```

So the native side has the dispatch primitives. The View base class has `on_hover_enter` / `on_hover_leave` callbacks. Native `TextButton` / `HyperlinkButton` / `ShapeButton` all override `on_mouse_enter` / `on_mouse_leave` to flip an internal `hovered_` flag (`core/view/src/buttons.cpp:35-36, 59-60, 105-106, 124-125`) and re-paint themselves accordingly.

### `@pulp/react` prop-applier never calls `registerHover`
`packages/pulp-react/src/prop-applier.ts:73-79` (post-#1050 import):

```ts
function applyEventHandler(id: string, key: string, value: unknown): void {
    if (typeof value !== 'function') return;
    const eventName = eventNameFor(key);
    call('on', id, eventName, (...a: unknown[]) => (value as (...x: unknown[]) => void)(...a));
}
```

When `key === 'onMouseEnter'`, this registers a listener on `'mouseenter'` via the `on()` global — but doesn't call `registerHover(id)`. So the native dispatch is never armed for that widget. Listener exists, dispatch never fires, handler never runs.

This is a 2-line fix: detect `onMouseEnter` / `onMouseLeave` (or `onPointerEnter` etc.) in `applyEventHandler` and call `registerHover(id)` once per widget that has either handler.

### Spectr's actual hover patterns

Spectr's WebView source uses CSS `:hover` in 90%+ cases — those rules live in `<style>` blocks inside `editor.html` (or as CSS-in-JS strings). The native side has no CSS engine path that recognizes pseudo-classes.

Two places in Spectr's bottom-toolbar source use `onMouseEnter` / `onMouseLeave` instead (`spectr-editor-extracted.js:2263-2264`):

```jsx
onMouseEnter={e => { if (!disabled) e.currentTarget.style.background = 'rgba(120,180,255,0.14)'; }}
onMouseLeave={e => { e.currentTarget.style.background = 'transparent'; }}
```

This pattern works in WebView (React forwards to a real DOM `<button>` whose `style` is mutable). In native, `e.currentTarget.style` doesn't exist — even if the event ever fired, the imperative style mutation would silently no-op against the synthetic event-target.

### (a) Auto-register hover when `onMouseEnter` / `onMouseLeave` is set
P0 for any UI that uses React-style hover handlers. 2-line fix in prop-applier.ts. Also covers `onPointerEnter` / `onPointerLeave` if we want pointer-event parity.

### (b) Surface CSS `:hover` to native somehow
Two options:
1. **Parser-side translation** — `@pulp/css-adapt` (or web-compat-style-decl.js) parses `<style>` blocks, hoists `:hover` rules into per-component `_hoverStyle` props, and the bridge applies/reverts them on `mouseenter` / `mouseleave`. Requires a CSS parser path.
2. **Bypass — instruct consumers to switch to `onMouseEnter`** in design-import output. Doesn't help Spectr's existing extracted source but is cheap and easier than a full pseudo-class engine.

Option 2 unblocks `pulp import-design`-emitted apps but leaves any hand-extracted code (Spectr) hanging. Option 1 is the long-term fix; option 2 is a fast unblock.

### (c) `e.currentTarget.style.X = Y` synthetic event-target
Lower priority — only matters for code that relies on imperative DOM mutation in event handlers, not React-state-driven re-renders. Could be addressed by making the synthetic event target's `.style` proxy through to bridge `setBackground` / `setBorder` etc. — non-trivial. Or just deprecate the pattern in design-import output.

For Spectr specifically, fixing (a) + telling the few onMouseEnter consumers to use React state instead of imperative `.style` mutation is the path to parity.

## Tap-state status (related)

React state-based tap-flash patterns (`[flash, setFlash]` + `setTimeout` in `RailBtn`) are framework-agnostic and DO work post-#1073 — verified for CLEAR button in the Spectr audit (`planning/audit-2026-04-30-native-vs-webview.md`). They DON'T work for items inside popovers because the click never dispatches there (#1148). So tap states aren't a separate framework gap — they're a downstream consequence of #1148 staying open.

CSS `:active` is the same shape as `:hover` above — needs (b)-style support if we want it.

## Cross-refs

- Sister issues: **#1147** (popover render), **#1148** (overlay click dispatch) — together they form the "interaction parity gap" trio for Spectr's port.
- Umbrella: **#924**.
- The bridge primitive (`registerHover`) was likely added during early bridge work — pre-pulp-react import (#1042/#1050). Just needs wiring in the prop-applier today.

## Spectr-side artifacts

- Audit (46 screenshots): https://github.com/danielraffel/spectr/blob/feature/native-react-editor/planning/audit-2026-04-30-native-vs-webview.md
- Tested SDK: v0.68.0
- Spectr branch: `feature/native-react-editor` @ `7715a24`

danielraffel:
The native bridge has had `registerHover(id)` since da48acf6 (initial
hover-callback work pre-#652 import) and the JS preamble at
core/view/src/widget_bridge.cpp:483-518 even auto-arms it from any
`on(id, 'mouseenter'|'mouseleave', fn)` call (added in df11a88b /
#1008). But @pulp/react's prop-applier left two real gaps that kept
hover dispatch dormant for typed React consumers:

  1. `onPointerEnter` / `onPointerLeave` registered listeners under
     'pointerenter' / 'pointerleave' event keys. The native dispatch
     only fires 'mouseenter' / 'mouseleave' though, so those handlers
     never ran. Now `eventNameFor` aliases the pointer-shaped names
     to the mouse event names so the listener key matches what
     __dispatch__ sends.

  2. The bridge JS preamble's auto-arm relies on `on()` being called
     with the right event name, but that's wired through the bridge's
     C++ string preamble — mock-bridge tests, embedded harnesses, or
     any older runtime without the #1008 preamble drop the auto-arm.
     `applyEventHandler` now calls `registerHover(id)` directly when
     it sees any of the four hover-shaped React props, deduped via
     a module-scope `hoverArmed` Set. Native registerHover is
     idempotent (assignment-rebind on the View's callback slots), so
     the redundant call when the preamble auto-arms is a no-op.

Cleanup is recursive on subtree removal: React's commitDeletion only
calls removeChild on the root of a deleted subtree, but the native
removeWidget walks the full subtree on the C++ side. host-config now
maintains an `instanceById` registry and `clearHoverForSubtree` walks
the children so re-mounts that re-use a former descendant id re-arm
cleanly. (Codex P1 review caught this; would have leaked stale
hoverArmed entries otherwise.)

Public types: BaseProps now extends a new HoverProps interface so
typed consumers writing `<Button onMouseEnter={...} />` or
`<View onPointerEnter={...} />` type-check correctly. HoverProps is
re-exported from index.ts.

Tests cover: arm-once on mount, no-arm without hover handler, dedup
across the four hover props, dedup across mount + applyChangedProps,
re-arm via applyChangedProps when the handler is added later, the
pointer→mouse event-name aliasing, the mouseenter+pointerenter
collapse-to-single-listener path, clearHoverFor + re-mount, and the
recursive subtree cleanup (build → arm leaves → removeChild on the
root → re-mount under a former descendant id → re-arm fires).

Codex consult verdict: GOOD after addressing all P1s (idempotency,
pointer aliasing, recursive cleanup, public type exposure). 16 tests
pass; tsc + esbuild produce a clean 950kb ESM bundle. Spectr v0.68.0
build-out via /tmp/pulp-react/packages/pulp-react/pulp-react-0.0.1.tgz
confirms the wired bundle includes the hover code. Headless render
(planning/screenshots/pulp-1149-hover-demo.png) shows registerHover
fires once for a button with onMouseEnter and not at all for a button
with only onClick.

No `unregisterHover` primitive exists on the bridge, but it's not
needed: removeWidget destroys the View and its on_hover_enter /
on_hover_leave closures atomically. Filed as a follow-up only if a
future widget API needs to detach hover without destroying the view.

Closes #1149.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.68.0
+    VERSION 0.68.1
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/packages/pulp-react/src/bridge.ts
+++ b/packages/pulp-react/src/bridge.ts
@@ -108,6 +108,14 @@ declare global {
     // ── Theme ───────────────────────────────────────────────────────
     function setTheme(name: 'dark' | 'light' | 'pro_audio' | string): void;
 
+    // ── Event opt-in registrars ─────────────────────────────────────
+    /// Arm `mouseenter` / `mouseleave` dispatch on a widget. Must be
+    /// called once per widget that has an `onMouseEnter` / `onMouseLeave`
+    /// (or `onPointerEnter` / `onPointerLeave`) handler — otherwise the
+    /// listener registered via `on(id, 'mouseenter', fn)` stays dormant.
+    /// pulp #1149.
+    function registerHover(id: string): void;
+
     // ── Layout flush + frame service ────────────────────────────────
     /// Force a layout pass on the root container. Used in
     /// `resetAfterCommit` so the host config owns commit-time flush.
@@ -144,7 +152,7 @@ export function createMockBridge(): MockBridge {
         'setBorderSide', 'setOpacity', 'setVisible', 'setPosition',
         'setText', 'setTextColor', 'setTextAlign',
         'setSpectrumData', 'setWaveformData', 'setMeterLevel', 'setProgress',
-        'setValue', 'setTheme', 'layout', 'on',
+        'setValue', 'setTheme', 'layout', 'on', 'registerHover',
     ];
     const saved: Record<string, unknown> = {};
     return {
--- a/packages/pulp-react/src/host-config.ts
+++ b/packages/pulp-react/src/host-config.ts
@@ -18,7 +18,7 @@ import type {
     PulpContainer,
     IntrinsicElementName,
 } from './types.js';
-import { applyAllProps, applyChangedProps } from './prop-applier.js';
+import { applyAllProps, applyChangedProps, clearHoverFor } from './prop-applier.js';
 
 type Type = IntrinsicElementName;
 type Props = Record<string, unknown>;
@@ -173,7 +173,7 @@ export const PulpHostConfig: HostConfig<
         // (Codex + RepoPrompt review both flagged the original plan that
         // tried createX(id, parent) here as the wrong lifecycle point.)
         const id = (props.id as string) ?? autoId(rootContainer);
-        return {
+        const inst: Instance = {
             id,
             type,
             // parentId stays undefined until attached
@@ -182,6 +182,10 @@ export const PulpHostConfig: HostConfig<
             onBridge: false,
             pendingChildren: [],
         };
+        // Track the instance so removeChild can walk its subtree
+        // and clear hover bookkeeping for descendants (pulp #1149).
+        instanceById.set(id, inst);
+        return inst;
     },
 
     createTextInstance(
@@ -204,14 +208,16 @@ export const PulpHostConfig: HostConfig<
         // reconciler treats it as an opaque host-text type, but our
         // appendChild path will materialize it via createLabel + setText
         // when its parent attaches.
-        return {
+        const inst: Instance = {
             id,
             type: 'Label',
             props: { children: String(text), text: String(text) },
             childIds: [],
             onBridge: false,
             pendingChildren: [],
-        } as unknown as TextInstance;
+        };
+        instanceById.set(id, inst);
+        return inst as unknown as TextInstance;
     },
 
     shouldSetTextContent(type, _props) {
@@ -271,6 +277,10 @@ export const PulpHostConfig: HostConfig<
     removeChildFromContainer(_container, child) {
         // Detach from root and let the bridge clean up the subtree.
         if (typeof g.removeWidget === 'function') call('removeWidget', child.id);
+        // Mirror detach()'s hover bookkeeping cleanup (pulp #1149) —
+        // recursive because removeWidget destroys the whole subtree on
+        // the native side.
+        clearHoverForSubtree(child);
     },
 
     clearContainer(_container) {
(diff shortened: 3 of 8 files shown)


danielraffel:
Closing — superseded by #1173 (your simpler version, just merged). Note: this PR included a recursive subtree-cleanup walker for hover state on unmount (Codex P1 fix). If hover-state leaks become observable in practice, can port that walker as a follow-up PR.

PR #1172, opened 2026-04-30
fix(view): split web-compat fontFamily list and resolve each candidate (#1151)

### LINKED ISSUE #1151: core/view: web-compat fontFamily passes whole comma-separated list to setFontFamily — splits/fallback walk not implemented
## Summary

`web-compat-style-decl.js:465` (the web-compat → bridge style adapter) currently passes the **entire comma-separated CSS `font-family` list** to `setFontFamily` after only stripping outer quotes. That means a CSS value like:

```css
font-family: 'JetBrains Mono', ui-monospace, SFMono-Regular, Menlo, Consolas, monospace;
```

…arrives at the bridge as the literal single string `JetBrains Mono, ui-monospace, SFMono-Regular, Menlo, Consolas, monospace` (or similar after quote-strip). `SkFontMgr::matchFamilyStyle()` then never matches anything, because that's not a registered family name — it's a list.

Result: even when one of the listed fallback families *would* match (e.g. system has Menlo, or Pulp's bundled Inter+JBM matches the first entry), the lookup silently misses and we land on the platform default font. This is one of the contributing factors to typography drift visible in #1070.

## Where this lives

- `core/view/js/web-compat-style-decl.js:465` — the parser strips matching outer single/double quotes from the value but doesn't split on commas or normalize whitespace. Whatever string the CSS author wrote between the `:` and `;` (minus the quotes) becomes the family string the bridge sees.

This is independent of the bundled-font-registration gap filed separately (the new font-registration issue): even after that lands, this parsing bug will keep typography lookups silently missing.

## Repro

Any plugin (or design import) that uses CSS-in-JS or extracted CSS with the standard "preferred font, then a fallback list" pattern will hit this:

```jsx
<div style={{ fontFamily: "'Inter', system-ui, sans-serif" }}>hello</div>
```

In the WebView build: browser parses the list, walks through `Inter` → `system-ui` → `sans-serif`, finds first match, renders.

In native v0.68.0: the bridge gets the literal string `Inter, system-ui, sans-serif`, `matchFamilyStyle` returns null, falls back to default.

Spectr's source uses this pattern via `var(--mono)` / `var(--sans)` resolving to comma-separated lists; that's most of why typography looks drift-y on the Spectr port even on widgets where `fontSize` and `letterSpacing` are clearly applying correctly.

## Proposed fix

Standard CSS `font-family` parsing:
1. Split the value on top-level commas (don't split inside quoted segments — `"Foo, Bar"` is one family name; `Foo, Bar` is two).
2. Strip outer quotes from each segment.
3. Trim whitespace from each segment.
4. Try each family name in order via `matchFamilyStyle` (or whatever resolver path the bridge has). First successful match wins.
5. Generic family keywords (`serif`, `sans-serif`, `monospace`, `system-ui`, `ui-monospace`, `cursive`, `fantasy`) should map to platform-appropriate defaults (e.g. `monospace` → CoreText's monospace generic on macOS).

This is ~30 lines of JS; the resolver-loop bit is the only non-trivial piece (cycling through candidates against the typeface manager).

## Cross-refs

- Sister issue (just filed): font-registration API for plugins to provide their own .ttf — fixing both is needed for typography-fidelity parity.
- Tracked under umbrella #924 (Spectr-driven framework gaps).
- Related symptom: #1070 (PRESETS dropdown letter-spacing — but the family-resolution miss may be a contributing factor since the wrong font has a different intrinsic letter-spacing).

danielraffel:
web-compat-style-decl.js:465 stripped outer quotes but passed the entire
comma-separated CSS font-family value to setFontFamily as one string.
SkFontMgr::matchFamilyStyle never matched, so any plugin using a
fallback chain (e.g. `'JetBrains Mono', ui-monospace, monospace`) drifted
to the platform default font even when a listed candidate was bundled or
installed.

This change:

- adds `_parseFontFamilyList` to css-parser.js that splits on top-level
  commas (skipping commas inside single- or double-quoted segments),
  trims whitespace and strips outer quotes from each candidate. The
  candidates are joined with `;` and handed to the existing
  setFontFamily bridge call — no new bridge primitive.

- updates Skia `get_cached_typeface` to walk the `;`-joined chain and
  return the first typeface that resolves via the existing bundled +
  matchFamilyStyle path. When every named candidate misses, it falls
  through to `matchFamilyStyle(nullptr, ...)` for the platform default,
  preserving the prior single-family "always render something" behaviour.

- updates the macOS CoreGraphics path to walk the same chain. Because
  CTFontCreateWithName silently substitutes a fallback face for an
  uninstalled name, each candidate is first probed via
  CTFontDescriptorCreateMatchingFontDescriptor and a case-insensitive
  family-name comparison so missing candidates don't short-circuit the
  walk to a substituted face (Codex P1 on the v1 patch).

Tests added in the same PR (`[issue-1151]`):
- 10 JS-side parser tests (test_web_compat_prelude.cpp): comma list,
  quoted-with-internal-comma, single name, leading whitespace,
  generic-only, empty string, double-quoted comma family, drop-empty.
- 3 widget-bridge integration tests (test_widget_bridge.cpp): assert
  the bridge stores the joined chain on `Label::font_family()`.
- 1 CoreGraphics chain test (test_canvas.cpp): walk past a missing
  first candidate and still render text.
- 3 Skia chain tests (gated behind `#ifdef PULP_HAS_SKIA`): mid-chain
  match equivalence with direct lookup, generic-only fallback,
  all-missing fallback to platform default.

All targeted tests pass locally; existing 87 widget-bridge and 71
web-compat-prelude tests still green.

Closes #1151.

### DIFF
--- a/.agents/skills/engine/SKILL.md
+++ b/.agents/skills/engine/SKILL.md
@@ -145,3 +145,16 @@ Don't refactor this into a "pure native" shape — there's no way to do it witho
 `setTimeout(fn, 0)` deliberately bypasses `__scheduleTimer__` and routes through `Promise.resolve().then(...)` so it drains on the next `pump_message_loop()` call. This matches React's scheduler expectations and makes tests deterministic (no host frame loop needed). Positive-delay timeouts go through the native deadline tracker and only fire when `service_frame_callbacks()` runs.
 
 If a consumer reports "my setTimeout(fn, 1) never fires", check that the host is actually calling `service_frame_callbacks()` from its frame loop — that's the drain hook for non-zero delays.
+
+### CSS font-family lists are joined with `;` across the JS↔C++ boundary (#1151)
+
+`web-compat-style-decl.js` parses a comma-separated CSS `font-family` value into an ordered candidate list, joins it with `;`, and hands the joined string to the existing `setFontFamily(id, family)` bridge call. The C++ side (`get_cached_typeface` in `core/canvas/src/skia_canvas.cpp` and `create_font_with_fallback` in `core/canvas/platform/mac/cg_canvas.mm`) splits on `;` and walks the chain.
+
+We use `;` (not `,`) because:
+
+1. `,` is a legal character inside CSS quoted family names (`"Foo, Bar"` is one family) and the C++ bridge has no way to know which commas were "list separators" vs "literal name characters" once the string crosses the boundary. The JS parser is the only layer that has full quoting context.
+2. `;` is *not* legal anywhere inside an unquoted CSS family name (per CSS Fonts grammar), so it's a safe delimiter.
+
+If you ever change the delimiter, change both sides in lockstep — the cache key in `get_cached_typeface` hashes the joined string verbatim, so a mismatch silently bypasses the cache and may also bypass the chain-walk entirely.
+
+The CG path additionally uses `CTFontDescriptorCreateMatchingFontDescriptor` to verify each candidate is *actually installed* before accepting it, because `CTFontCreateWithName` silently substitutes the system default for a missing name — that would short-circuit the chain walk to whatever CoreText guesses is closest.
--- a/core/canvas/platform/mac/cg_canvas.mm
+++ b/core/canvas/platform/mac/cg_canvas.mm
@@ -2,6 +2,9 @@
 
 #ifdef __APPLE__
 
+#include <cctype>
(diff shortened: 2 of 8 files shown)


danielraffel:
Closing — DIRTY against main and superseded by your #1174 which has the cleaner variadic-args bridge contract.

PR #1190, opened 2026-05-01
test(canvas): cover path to sdf guard edges

danielraffel:


danielraffel:
Follow-up pushed at 69aa0360 after the first Windows Namespace run exposed a hot-reload watcher timing failure outside the path-to-SDF tranche. The branch is now rebased on current main, hot-reload watcher tests use deterministic temp paths/content waits plus a CTest resource lock, and focused local validation passed: pulp-test-hot-reload, HotReloader CTest, pulp-test-path-to-sdf, path_to_sdf CTest, skill-sync/version-bump reports, and git diff --check. Fresh PR-event checks are queued.

PR #1212, opened 2026-05-01
test(tools): cover iwyu annotate edges

danielraffel:


danielraffel:
Phase 3 Codecov coverage tranche for #643; tracked from #641. Focused target: tools/scripts/iwyu_annotate.py.

PR #1216, opened 2026-05-01
test(tools): cover status ladder edge paths

danielraffel:


danielraffel:
Phase 3 Codecov coverage tranche for #643; tracked from #641. Focused target: tools/check_status_ladder.py.

PR #1236, opened 2026-05-02
test(ci): cover cobertura merge edges

danielraffel:

PR #1239, opened 2026-05-02
test(runtime): cover websocket frame edge paths

danielraffel:


danielraffel:
Phase 3 Codecov tranche for #641.

Local validation before opening:
- `cmake --build build --target pulp-test-websocket-channel`
- `ctest --test-dir build -R "WebSocket" --output-on-failure` (11/11)
- `./build/test/pulp-test-websocket-channel` (62 assertions / 11 test cases)
- earlier skill-sync/version-bump and diff checks

Rebased onto current `origin/main` after #1236 merged, then opened through `shipyard pr --skip-target mac --skip-target ubuntu --skip-target windows`; the no-local-target exit is expected for this Namespace-only route.

Issue #1148, opened 2026-04-30
@pulp/react: clicks don't dispatch inside absolutely-positioned overlays + no outside-click/ESC close (popover interaction gap, post-#1073)


danielraffel:
## Summary

Spectr v0.68.0 (Pulp v0.68.0 SDK, after #1073 fixed top-level click bubbling) ships with a new interaction-parity gap: clicks **inside** absolutely-positioned overlay panels (popovers, dropdowns, menus) do not dispatch React `onClick` handlers, **outside-click** does not auto-close them, and **ESC** does not close them. The only way to close any open menu is to re-tap its trigger button.

## Repro

Spectr's bottom toolbar has multiple dropdown triggers (CLEAR, ⋯, SCULPT, PEAK, PRESETS, BANDS). Each opens a `position: absolute; bottom: 34; left: 0` overlay panel with `<button onClick={…}>` items inside.

Tested behavior on v0.68.0:
| Action | Expected | Actual |
|--------|----------|--------|
| Tap BANDS trigger | Open dropdown | Opens ✓ |
| Tap "48" inside BANDS dropdown | Select 48, close | **Nothing happens** — no visible focus, no selection, no close |
| Tap SCULPT trigger to open EDIT MODE | Open dropdown | Opens ✓ |
| Tap LEVEL row inside EDIT MODE | Set mode, close | **Nothing happens** |
| Tap on white canvas behind dropdown | Close dropdown | **Stays open** |
| Press ESC with dropdown open | Close dropdown | **Stays open** |
| Tap original trigger again | Close dropdown | Closes ✓ (toggle behavior) |

The trigger button toggles via `onClick={() => toggleMenu(k)}` and that *does* reach @pulp/react after #1073. The issue is purely with clicks landing on **descendants of the absolutely-positioned panel**.

## Why this is distinct from #1067/#1073

#1067/#1073 fixed click events on widgets in the **regular layout flow** (e.g., the trigger buttons themselves). The new gap is specifically with the **overlay subtree**:

1. The popover is `position: absolute` so its widget rect is computed relative to the relative ancestor — Pulp's hit-test stack may resolve a global click coordinate to the FilterBank canvas underneath it instead of the popover, because the canvas is laid out in the document flow at full plugin width while the popover's widget might not be registered as "above" the canvas in z-order.
2. Or: clicks inside the popover *do* land on the right widget but the widget's children (the `<button>` rows) lose their click registrar when re-rendered as overlay descendants.

[...]

Issue #1147, opened 2026-04-30
@pulp/react: nested flex content + inline SVG don't render inside absolutely-positioned popovers (PEAK/SCULPT empty + content overflow on v0.68.0)


danielraffel:
## Summary

In Spectr v0.68.0 (built on Pulp v0.68.0 SDK), opening any flexbox popover renders **only the panel background and header**; the inner button rows lose **all child content** — SVG icons, label `<span>`s, description `<span>`s. With certain layouts the description text escapes the panel and paints to the right on whatever background is behind the popover.

This is a parity-blocker for any UI that uses common React patterns like `<button><svg/><span>{label}<span>{desc}</span></span></button>` inside an `position: absolute; bottom: …` overlay.

## Reproductions

Two popovers in Spectr's bottom toolbar. Both are `position: absolute; bottom: 34; left: 0; width: 280; padding: 6; display: flex; flexDirection: column; gap: 2; background: rgba(12,16,22,.96)`. Their items are `<button>`s containing inline `<svg><path/></svg>` plus nested `<span>`s.

### Bug 1 — PEAK / ANALYZER popover: content escapes panel
[`spectr-editor-extracted.js:3177`](https://github.com/danielraffel/spectr/blob/main/native-react/spectr-editor-extracted.js) (`AnalyzerPopover`)

Each item is structured:

```jsx
<button style={{ ...display: 'block', textAlign: 'left' }}>
  <span style={{ display: 'flex', alignItems: 'center', gap: 8 }}>
    <span style={{ display: 'inline-block', width: 18, height: 2, background: a.color }} />
    <span style={{ letterSpacing: 1.5, fontWeight: 600 }}>{a.label}</span>   {/* PEAK */}
  </span>
  <span style={{ display: 'block', fontSize: 9.5, opacity: 0.6, marginTop: 3 }}>
    {a.desc}                                                                  {/* description */}
  </span>
</button>
```

What renders:
- Panel + header `ANALYZER · A to cycle` ✓
- Active row's button background ✓ (only)
- Description spans (`a.desc`) — render at correct y but **far to the right of the panel**, overlapping the canvas behind the popover
- Color pill `<span inline-block width:18 height:2>` — missing
- Label spans (`PEAK`, `AVG`, `BOTH`, `OFF`) — missing

Screenshot: https://github.com/danielraffel/spectr/blob/feature/native-react-editor/planning/screenshots/native-editor-v0.68.0-PEAK-dropdown-content-overflow.png

[...]

Issue #1070, opened 2026-04-30
view: PRESETS dropdown letter-spacing not honored vs WebView (post-v0.66.0)


danielraffel:
## Repro

- Branch: `main` at SDK v0.66.0 (after #1006/#1008 click-dispatch fix and #964 canvas wiring)
- Build: Spectr standalone, `-DSPECTR_NATIVE_EDITOR=ON` (native-bridge path), macOS arm64
- Reference: Spectr standalone, `-DSPECTR_NATIVE_EDITOR=OFF` (WebView path), same SDK
- Action: open the bottom-toolbar PRESETS dropdown
- Fixture: Spectr feature/native-react-editor at the v0.66.0 SDK pin

## Symptom

The PRESETS dropdown's `FACTORY` section header (CSS: `letterSpacing: 2`, fontSize 8.5, opacity 0.4) renders with letters visibly tight together in the native bridge, while the WebView reference renders the same JSX with the expected wide tracking ("F A C T O R Y" with clear gaps).

The same regression is visible on the menu-item text below it (`HARMONIC SERIES`, `ALTERNATING`, `VOCAL FORMANTS`, `SUB ONLY (< 160 Hz)`, etc.). React menu-item style is `letterSpacing: 0.8` (`menuItem` in `native-react/dist/editor.js`); the WebView shows the wider tracking, the native bridge does not.

## Reference screenshots

- Native: https://github.com/danielraffel/pulp-planning/blob/399bb3c1c10d1a2b528ff59d75a40066d2534770/screenshots/v0.66.0-dropdown-native.png
- WebView: https://github.com/danielraffel/pulp-planning/blob/399bb3c1c10d1a2b528ff59d75a40066d2534770/screenshots/v0.66.0-dropdown-webview-open.png

Crop comparison ("FACTORY" header, same scale):

| Native | WebView |
|---|---|
| Letters visually adjacent, no `letter-spacing: 2` tracking | Clear inter-letter gaps consistent with `letter-spacing: 2` |

## Likely framework gap

`@pulp/css-adapt` is forwarding `letterSpacing` to the bridge (verified in `native-react/dom-adapter.tsx` line 119, 402, 798) and the chrome-layer Label calls do honor it after #927 / #969. The dropdown text is rendered as JSX text inside React `<button>` and `<div>` elements, which the dom-adapter wraps as Labels — these Labels' `setLetterSpacing` calls fire but the rasterized output does not advance characters by the requested amount. Likely candidates:

[...]

Issue #998, opened 2026-04-29
view: v0.61.0 regression — position:absolute; inset:0 no longer fills parent (FilterBank canvas shrinks to bottom strip)


danielraffel:
## Repro (v0.60.0 → v0.61.0 regression, deterministic)

Built Spectr's native-bridge editor against v0.60.0 → working layout (FilterBank canvas fills the editor center). Bumped pin to v0.61.0 → **FilterBank canvas widget is now sized to ~5% of the parent (a bottom strip), not full-bounds**.

### Reproduction

Spectr branch: `feature/native-react-editor` HEAD `0897c38`. Two builds:

```
# v0.60.0 baseline
find_package(Pulp 0.60.0 REQUIRED)
cmake -B build -DSPECTR_NATIVE_EDITOR=ON
cmake --build build --target Spectr_Standalone
open ./build/Spectr.app
# → FilterBank fills editor center (visible as dark area below toolbar)

# v0.61.0 bump
find_package(Pulp 0.61.0 REQUIRED)
cmake -B build -DPulp_DIR=$HOME/.pulp/sdk/0.61.0/lib/cmake/Pulp -DSPECTR_NATIVE_EDITOR=ON
cmake --build build --target Spectr_Standalone
open ./build/Spectr.app
# → FilterBank shrinks to a strip at the very bottom of the editor
```

### Visual proof via probe

Set `<canvas>` bg to bright magenta in dom-adapter (one-line):
```ts
if (tag === 'canvas' && adapted.background === undefined) {
    adapted.background = '#ff00ff';
}
```

v0.61.0 result: bright magenta strip ONLY in the bottom ~30px of the window. The entire region between the top toolbar and that strip is the App's parent (white from the standalone window's NSView default).

## What changed

The FilterBank canvas widget JSX is:

```jsx
// extracted.js:2155
<div ref={wrapRef} style={{ position: 'absolute', inset: 0, ... }}>
  <canvas ref={canvasRef} style={{ position: 'absolute', inset: 0 }} />
  <canvas ref={overlayRef} style={{ position: 'absolute', inset: 0, pointerEvents: 'none' }} />
</div>
```

In v0.60.0, `position: absolute; inset: 0` filled the parent (1320×~720 area between toolbars). In v0.61.0, it lays out as a small bottom strip.

dom-adapter's stretch-Y logic for absolute+top+bottom→height=860 is unchanged. So either:
[...]

PR #1293, opened 2026-05-03
fix(view): defer first @pulp/react render() to microtask (P0, closes #1292)

### LINKED ISSUE #1292: @pulp/react: intermittent crash on App() — 'cannot read property useState of null' (React dispatcher not yet bound)
## Summary

Spectr's native build (`feature/native-react-editor` @ `5f45d4a`, SDK pin v0.69.1) crashes during `WidgetBridge::load_script` with `std::runtime_error: failed to evaluate user_script: cannot read property 'useState' of null` thrown from inside React's reconciler-init path (line :629 of the bundled editor.js, called as `App` body at :29415).

This was intermittent — the *first* launch in a fresh session captured a working screenshot showing FilterBank spectrum + full toolbar render. The next 3+ launches in the same session all crashed at the same point with the same error. Reproduces against **both v0.69.0 and v0.69.1** SDKs (v0.68.1 not retested), so it is **not** a v0.69.1 regression — it's pre-existing and got newly visible.

## Repro

```bash
cd /Users/danielraffel/Code/spectr
cmake -S . -B build -DCMAKE_BUILD_TYPE=Release \
  -DPulp_DIR=$HOME/.pulp/sdk/0.69.1/lib/cmake/Pulp \
  -DSPECTR_NATIVE_EDITOR=ON
cmake --build build --target Spectr_Standalone -j8
./build/Spectr.app/Contents/MacOS/Spectr
```

## Stack trace (JS-side, captured from stderr)

```
libc++abi: terminating due to uncaught exception of type std::runtime_error:
failed to evaluate user_script: cannot read property 'useState' of null
    at <anonymous> (:629)
    at App (:29415)
    at He (:4800)         ← React.createElement (dist-name-mangled)
    at ai (:7657)
    at $h (:7353)
    at Zh (:7347)
    at Ph (:7332)
    at Lh (:7123)
    at ad (:4084)
    at af (:6944)
    at <anonymous> (:8284)
    at render (:24121)
    at render ()
    at <anonymous> (:29585)
    at <eval> (:29586)
```

## C++-side stack (from `~/Library/Logs/DiagnosticReports/Spectr-…ips`)

```
[ 0] Spectr  pulp::view::eval_or_throw(ScriptEngine&, char const*, std::string const&)
[ 0] Spectr  pulp::view::WidgetBridge::load_script(std::string const&)
[ 0] Spectr  spectr::NativeEditorView::NativeEditorView(spectr::Spectr&)
[ 0] Spectr  spectr::Spectr::create_view()
[ 0] Spectr  pulp::format::ViewBridge::open(std::string*)
[ 0] Spectr  pulp::format::StandaloneApp::run_with_editor(bool)
[ 0] Spectr  main
```

JS log immediately preceding the crash:

```
[spectr-js] SHIM_LOAD: __spectrLog wired ok
[spectr-js] shim: rAF was=function existing=true reqFrame=function
[spectr-js] [jsx] #1 <App>
[spectr-js] [hc#1] layout()  ← host-config layout() called once
libc++abi: terminating ...
```

## Hypothesis

`React.useState` resolves to `ReactCurrentDispatcher.current.useState` at runtime. If `ReactCurrentDispatcher.current` is null when App() runs, the `.useState` access throws "cannot read property of null". The dispatcher is set by the renderer (`reconciler.updateContainer` → render path) — so this happens when *App() is called outside a render commit*, OR when the renderer's "current" pointer hasn't been bound to the right dispatcher.

`@pulp/react` calls `reconciler.updateContainer(element, fiberRoot, null, null)` inside `render()`. If the reconciler internal state isn't ready when that fires (e.g. the host config returned early from a layout call), App() can be invoked in a "bare" context without a dispatcher.

The fact that **the first launch worked** strongly suggests a race/timing issue: probably between the JS-engine init, the bridge-globals registration, and the first reconciler render pass. Cold heap state may differ between launches.

## Impact

P0 — **completely blocks Spectr's native build from launching** when the timing race fires. Visible to anyone running the standalone today.

## Suggested fixes

1. **Defer App() to a microtask** — wrap the user's `render(<App />, root)` call so it runs after a `Promise.resolve().then(...)`. Gives the reconciler one tick to fully bind dispatchers.
2. **Eager dispatcher install** — in `host-config.ts`, ensure `reconciler.injectIntoDevTools` (or whichever React-internal hook is needed) runs before any `updateContainer`.
3. **Guard `<App>` rendering** — verify React.version + dispatcher presence inside `render()` and throw a clearer error if not ready.

A vitest case can probably reproduce this by calling `render(<App/>)` immediately after `createRoot()` without any setTimeout, but only when the bridge globals are still being installed asynchronously.

## Cross-refs

- Spectr branch: `feature/native-react-editor` @ `5f45d4a`
- Last working bundle: `native-react/dist/editor.js` (Apr 30 18:59, 1.25 MB) — unchanged across the regression window
- Sister Spectr-blockers: #964, #994 (closing soon via #1291), #998, #1147, #1148, #1070

danielraffel:
## Summary

**P0** — closes #1292. Spectr's native build crashed intermittently during `WidgetBridge::load_script` with `cannot read property 'useState' of null` thrown from inside React's reconciler-init path when App() was invoked. Bug is a race between react-reconciler's internal dispatcher binding and the synchronous render path.

## Root cause

`reconciler.createContainer` leaves the dispatcher slot in a transitional state. When `updateContainer` fires immediately after, the function component's `useState()` call resolves the dispatcher through `ReactCurrentDispatcher.current` — which can still be null. The race is sensitive to cold heap state: first launch in a session frequently won the race; subsequent launches lost it.

## Fix

Per #1292's option 1 (recommended): defer the **first** `updateContainer` call to a microtask. Subsequent renders remain synchronous so test harnesses + AOT-bundled plugin code keep their "bridge calls flushed before render() returns" contract.

```ts
// First render: queueMicrotask(updateContainer)
// Subsequent renders: synchronous as before
if (isFirstRender) {
    queueMicrotask(() => reconciler.updateContainer(...));
    return c;
}
reconciler.updateContainer(...);
```

`queueMicrotask` is available on every modern JS engine (QuickJS, JSC, V8). Promise fallback included.

## Tests

4 new vitest cases in `packages/pulp-react/test/render-dispatcher-1292.test.ts`:
- First render emits no bridge calls synchronously
- Bridge calls land after one `Promise.resolve()` drain
- Subsequent renders are synchronous (no deferral)
- `React.useState()` inside App() does not crash on first render

All 21 vitest cases pass locally.

## Cross-refs

- Issue: #1292 (P0, blocks Spectr standalone)
- Sister Spectr-blockers: #964, #994 (closing via #1291), #998, #1147, #1148, #1070
- Reproduces against v0.69.0 + v0.69.1 — not a v0.69.1 regression

### DIFF
--- a/packages/pulp-react/src/index.ts
+++ b/packages/pulp-react/src/index.ts
@@ -54,14 +54,16 @@ let defaultContainer: PulpContainer | null = null;
 export function render(element: ReactElement, container?: PulpContainer): PulpContainer {
     const c = container ?? (defaultContainer ??= createRoot(''));
     let rec = rootsByContainer.get(c);
+    const isFirstRender = !rec;
     if (!rec) {
         const fiberRoot = reconciler.createContainer(
             c,
             // LegacyRoot = synchronous mode. Matches the v0 architecture
-            // doc ("Concurrent mode: deferred for v0") and means each
-            // render() returns after the bridge calls have all been
-            // emitted — no microtask gap, which is what tests and
-            // AOT-bundled plugin code both expect.
+            // doc ("Concurrent mode: deferred for v0"). After the first
+            // render the dispatcher is bound and subsequent updates
+            // remain synchronous — render() returns after the bridge
+            // calls have all been emitted, which is what tests and
+            // AOT-bundled plugin code expect.
             LegacyRoot,
             null,
             false,
@@ -73,6 +75,38 @@ export function render(element: ReactElement, container?: PulpContainer): PulpCo
         rec = { container: c, fiberRoot };
         rootsByContainer.set(c, rec);
     }
+
+    // pulp #1292 P0 — on the FIRST render, defer the initial
+    // updateContainer to a microtask. createContainer leaves the
+    // reconciler's internal dispatcher slot in a transitional state;
+    // calling updateContainer immediately can race the dispatcher
+    // binding, causing function components to invoke `useState` /
+    // `useEffect` before `ReactCurrentDispatcher.current` is set
+    // (intermittent on cold starts — first launch in a session
+    // happened to win the race, subsequent ones lost it).
+    //
+    // Yielding one microtask gives the reconciler a chance to fully
+    // bind dispatchers before App() runs. Subsequent renders go
+    // through the synchronous path so test harnesses + AOT plugin
+    // code see the bridge-calls-before-return guarantee they expect.
+    if (isFirstRender) {
+        const r = rec;
+        // Use queueMicrotask when available (every modern JS engine —
+        // QuickJS, JSC, V8) and fall back to Promise.resolve().then().
+        const enqueue: (fn: () => void) => void =
+            (typeof queueMicrotask === 'function')
+                ? queueMicrotask
+                : (fn) => { void Promise.resolve().then(fn); };
+        enqueue(() => {
+            try {
+                reconciler.updateContainer(element, r.fiberRoot, null, null);
+            } catch (err) {
+                console.error('[@pulp/react] initial updateContainer failed:', err);
+            }
+        });
+        return c;
+    }
+
     reconciler.updateContainer(element, rec.fiberRoot, null, null);
     return c;
 }
--- a/packages/pulp-react/test/render-dispatcher-1292.test.ts
+++ b/packages/pulp-react/test/render-dispatcher-1292.test.ts
@@ -1,0 +1,109 @@
+// pulp #1292 P0 — render() must not run App() before React's dispatcher
+// is bound, otherwise function components calling useState / useEffect
+// throw "cannot read property 'useState' of null" intermittently on
+// cold starts.
+//
+// Repro from the bug report: App() runs from inside the FIRST
+// updateContainer call. If the reconciler hasn't fully bound its
+// internal dispatcher before App() executes, the dispatcher slot is
+// null and `useState` blows up.
+//
+// Fix: defer the initial updateContainer to a microtask so the
+// reconciler finishes its dispatcher binding first.
+//
+// These tests verify the deferral happens on first render and that
+// subsequent renders stay synchronous.
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import React from 'react';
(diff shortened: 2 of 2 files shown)


danielraffel:
## Heads-up: posted alternative root-cause fix

This microtask deferral hides the symptom of #1292 but the underlying bug is **two independent React module instances inside `editor.js`** — esbuild can't dedup across @pulp/react's self-bundled `dist/index.mjs` boundary, so user code and the reconciler each get their own `React.__SECRET_INTERNALS_…ReactCurrentDispatcher` singleton.

Investigation + 5/5 repro + 6/6 fix-validation: see [#1292 comment](https://github.com/danielraffel/pulp/issues/1292#issuecomment-4365502118).

Alternative fix at [`fix/react-dispatcher-1292`](https://github.com/danielraffel/pulp/tree/fix/react-dispatcher-1292) — externalizes `react` / `react-reconciler` / `scheduler` from the pulp-react bundle (1-line `package.json` change, dist drops 953 KB → 27 KB). Will open a PR shortly.

Microtask deferral may still be worth keeping as defense-in-depth (catches any future timing race), but won't fix the duplicate-instance issues that ALSO show up post-mount: split `useContext`, `useRef` identity mismatches, error-boundary cross-talk, `Suspense` not seeing throws from the other React, etc.

Recommend either (a) close this in favor of the externals fix, or (b) merge both with the externals fix as the canonical close for #1292 and this as the additional safety net.

Feel free to reject — happy to defer to your judgment on which lands.

danielraffel:
Closing — superseded by #1295 (Spectr subagent's root-cause fix).

Their analysis showed the dispatcher-null crash is caused by **two independent React module instances** in the consumer bundle (editor.js), not by a timing race in updateContainer. My microtask-defer papered over the symptom but wouldn't have fixed the underlying duplicate-React problem (where state updates / contexts / refs across the two instances are silently isolated).

#1295's one-line fix to externalize `react`, `react-reconciler`, and `scheduler` in @pulp/react's `dist/index.mjs` build is the canonical fix.

Branch + worktree at `/Users/danielraffel/Code/pulp-1292` retained for now in case defense-in-depth is wanted post-#1295 merge; can delete after.

Issue #1067, opened 2026-04-30
view: clicks still don't dispatch React onClick handlers in v0.66.0 (after #1008)


danielraffel:
## Repro on v0.66.0

Built Spectr standalone against v0.66.0 (which includes #1001 mouseUp validate + #1008 auto-wire native registrar). Empirical click test via real CGEvents (`cliclick c:x,y`):

- ✅ App survives every click (#1001 fix verified — was crashing pre-v0.62.0)
- ❌ **No onClick handler fires.** No state changes, no `setOpenMenu` calls, no `fireStatus("CLEARED GAINS")` from clicking CLEAR, no popover opens from clicking "32 bands ▾" or PRESETS.

## What I verified

1. **dom-adapter forwards `onClick` props correctly.** Added a temporary probe; logs **17 onClick forwardings** during mount, all `tag=button target=View`.
2. **`@pulp/react`'s prop-applier has the right routing.**
   - `isEventHandler('onClick')` returns true
   - `eventNameFor('onClick')` returns `'click'`
   - `applyEventHandler` calls `call('on', id, 'click', wrapper)`
3. **`globalThis.on` is registered** — pointer-event registrations on `pr_3` (FilterBank wrap) appear correctly in stdout: `[pa#17] on(pr_3, pointerdown, ...)`, `pointermove`, `pointerup`, etc.

So the JS-side path (`<button onClick={...}>` → dom-adapter → @pulp/react → call('on', id, 'click', cb)) appears intact.

But after the user clicks (real CGEvent through `cliclick`), the click handler never invokes. The Pulp-side log shows only periodic canvas redraws and `[NativeEditorView::paint]`, no event-firing log lines.

## Spectr-side empirical findings

- ALL 17 buttons appear visually (CLEAR, SCULPT, PEAK, PRESETS, LIVE, PRECISION, IIR, FFT, HYBRID, 32-bands, etc.).
- Tapping them via cliclick at correct screen coords: nothing happens. App stays alive (no crash, see #1001), no popover opens, no status banner update, no segmented-control switch.
- Spectr's branch for repro: `feature/native-react-editor` HEAD (with #998 workaround `346eb64`).

## Hypothesis

After #1008 auto-wires `View::on_click` when JS calls `on(id, 'click', cb)`, the click event's path from Cocoa NSEvent → PulpView::mouseUp → `_dragTarget->on_click` → JS handler may have a missing link. Possible:

[...]

Issue #1290, opened 2026-05-03
tools/import-design: CLI main() at 0% diff-coverage — needs refactor or LCOV exclusion (blocks #1060)


danielraffel:
## Summary

`tools/import-design/pulp_import_design.cpp` is the CLI entry point for `pulp import-design`. The bulk of `main()` is argument parsing + side-effect orchestration (writes ui.js / tokens.json / classnames.json / package.json + dom-adapter scaffold + screenshot validation). Coverage of this file via the existing test harness is **0% on the diff** — the only existing tests (`test/test_cli_import_design.cpp`) are library-level (`extract_claude_classnames` / `serialize_…`) plus a couple of `--shellout` cases that don't drive the new flag combinations introduced by #1031, #1035, and #1039.

This is currently blocking PR #1060 (#1039 css-adapt wire) — diff-coverage sits at 72% (threshold 75%) and the only file dragging it down is `pulp_import_design.cpp` (0%). Same pattern will block every future #995 sub-issue PR that touches this file.

## Why fix-by-test is hard

- The CLI's behavior depends on subprocess env vars, build-time CMake defines, and filesystem layout. The shellout tests exec the built binary and observe exit code + stdout/stderr — but the env-var override I tried (`PULP_CSS_ADAPT_VENDOR_DIR`) didn't propagate via `pulp::platform::exec()` (no env-map argument), so I pulled the test (#1180).
- The `if (fs::exists(pkg_path)) return 1;` guard added in #1060 is provably correct (the production code is right) but the shellout test reported `r.exit_code == 0` on Linux/macOS even with a pre-seeded package.json. Either the test environment's `out_dir` resolution differs from the binary's, or the binary cache served an old build. Either way: tests for this file are flaky.

## Three options (pick one or combine)

1. **Refactor** — extract pure functions from `main()` (a `RunOptions` struct, a `run_import(opts) -> int` entry point) and unit-test the orchestration logic against a mock filesystem + mock writer. This is the most durable but ~half a day of work.

2. **LCOV exclusion** — add `// LCOV_EXCL_START` / `// LCOV_EXCL_STOP` markers around `main()` (or thread `tools/import-design/pulp_import_design.cpp` through `coverage_config.json`'s `diff_cover_excludes` pattern). CLI mains are a recognized coverage hole; this is the conventional dodge. Pairs naturally with #1058 (LCOV_EXCL → diff-cover propagation).

[...]

PR #1297, opened 2026-05-03
feat(view): generalize overlay-click routing for React popovers (closes pulp #1148)

### LINKED ISSUE #1148: @pulp/react: clicks don't dispatch inside absolutely-positioned overlays + no outside-click/ESC close (popover interaction gap, post-#1073)

danielraffel:
## Summary

Closes #1148. Generalizes the existing ComboBox overlay-click routing pattern (commit `41c05c35`, April 2026) into a reusable `View::active_overlay_` primitive, then exposes it through @pulp/react via a single `<View overlay>` boolean prop.

## Background

The ComboBox dropdown is a paint-only overlay queued via `View::overlay_queue()` — the dropdown ROWS aren't in the hit-test tree. Without a pre-`hit_test` bounds check in window-host `mouseDown`, clicks on dropdown items resolve to whichever sibling/ancestor sits at that pixel. The mac host (`window_host_mac.mm:347-374`) already had this check for ComboBox specifically (`pulp::view::ComboBox::active_popup_`).

React popovers built from `<View position="absolute">` + `<Button>` children get NONE of that — they fall through to `hit_test` and clicks land wherever the tree-z-order resolves. This is exactly the symptom Spectr's audit flagged: SCULPT/PEAK/PRESETS rows don't dispatch on inner click; the click sometimes leaks to the desktop app behind Spectr.

## What this PR adds

**C++ surface (`core/view/`)**:

```cpp
class View {
  static View* active_overlay_;            // generic equivalent of ComboBox::active_popup_
  void claim_overlay();                    // sets active_overlay_ = this
  void release_overlay();                  // clears if matches this
  bool overlay_contains(Point window) const;  // window-coord bounds via parent-chain walk
  ~View() override { release_overlay(); }  // safety
};
```

`window_host_mac.mm` mouseDown order:
1. Inspector intercept (unchanged)
2. ComboBox::active_popup_ (unchanged — pinned by regression test)
3. **NEW** — `View::active_overlay_` bounds check; route directly if hit
4. Tree `hit_test` (unchanged)

**JS bridge (`widget_bridge.cpp`)**:
- `claimOverlay(id)` / `releaseOverlay(id)` — global handlers

**@pulp/react (`packages/pulp-react/`)**:
- `<View overlay>` boolean prop in `ViewProps`
- prop-applier auto-calls `claimOverlay(id)` on mount when set, `releaseOverlay(id)` when unset / removed / unmount

## Tests added

C++ Catch2 (`test/test_overlay_routing.cpp`, tag `[issue-1148]`):
- defaults to nullptr
- claim sets, release this-only, swap behavior
- dtor releases (RAII safety)
- `overlay_contains` walks parent chain for absolute coords
- **ComboBox path untouched** (regression-pinned by existing `test_combo_dropdown.cpp`)

C++ Catch2 (`test/test_widget_bridge.cpp`):
- `claimOverlay`/`releaseOverlay` JS handlers drive `View::active_overlay_`

vitest (`packages/pulp-react/test/prop-applier-overlay.test.ts`):
- mount-claim, prop-absent no-op, false→true, true→false, prop-removal release, no-op on same-truthy

## Local validation (macOS, all green)

- pulp-cli build ✓
- pulp-test-overlay-routing: 7 cases / 18 assertions ✓
- pulp-test-combo-dropdown (regression): 8 cases / 25 assertions ✓ (ComboBox path verified untouched)
- pulp-test-widget-bridge: 95 cases / 702 assertions ✓
- pulp-test-view: 44 cases / 208 assertions ✓
- @pulp/react vitest: 33 tests / 5 files ✓

## Note on x11/win32

The original task brief mentioned mirroring into `window_host_x11.cpp` and `window_host_win32.cpp` but those don't exist on `origin/main` — only `mac/` and `ios/`. iOS is touch-driven and routes through `rootView->on_mouse_down(pos)` directly without `_dragTarget`/hit_test, so wiring the overlay short-circuit there would change semantics. Left iOS alone; future desktop hosts can pick up the API.

## Spectr-side follow-up

Spectr's `editor.js` popovers (built from `<View position="absolute">`) need to add `overlay` to those Views to opt in. After this PR ships in a release, Spectr's bundle re-emit will pick it up automatically; until then, the `<svg><path>` migration noted in the import-design skill applies similarly.

## Cross-refs

- Issue: #1148
- Pattern source: ComboBox routing at commit `41c05c35` (April 2026, pinned by `test/test_combo_dropdown.cpp`)
- Related: #1147 (popover render — separate from click dispatch), #1067/#1073 (top-level click bubble)
- Spectr-side: blocks SCULPT/PEAK/PRESETS popover inner clicks

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.71.0
+    VERSION 0.72.0
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/view/include/pulp/view/view.hpp
+++ b/core/view/include/pulp/view/view.hpp
@@ -23,7 +23,13 @@ class FrameClock;
 class View {
 public:
     View() = default;
-    virtual ~View() = default;
+    virtual ~View() {
+        // pulp #1148 — clear the overlay slot if this dying View holds it.
+        // Without this, an unmounted React popover leaves a dangling
+        // pointer that the platform window host would dereference on
+        // the next click.
+        if (active_overlay_ == this) active_overlay_ = nullptr;
+    }
 
     View(const View&) = delete;
     View& operator=(const View&) = delete;
@@ -405,6 +411,39 @@ public:
     static void set_inspector_mouse_hook(std::function<bool(const MouseEvent&)> hook);
     static bool call_inspector_key_hook(const KeyEvent& e);
     static bool call_inspector_mouse_hook(const MouseEvent& e);
+
+    // ── Generalized overlay-click routing (pulp #1148) ───────────────────
+    //
+    // ComboBox already uses a `static ComboBox* active_popup_` pointer so
+    // the platform window-host layer can short-circuit hit-testing for
+    // paint-only overlays (the dropdown items render OUTSIDE the
+    // ComboBox's hit_test bounds). React popovers built from
+    // `<View position="absolute">` + nested buttons hit the same problem
+    // but have no widget-specific shortcut, so clicks fall through to
+    // whatever sibling/ancestor pixel is under the overlay.
+    //
+    // `active_overlay_` is the per-View opt-in equivalent: any View can
+    // claim itself as the global click-eligible overlay. The window host
+    // checks this AFTER `ComboBox::active_popup_` (so the ComboBox path
+    // stays exact-as-was) and BEFORE the regular tree hit_test. If the
+    // click falls inside the overlay's window-rect, it routes to the
+    // overlay; otherwise the overlay is auto-released and the click
+    // continues to the tree.
+    //
+    // The @pulp/react host config calls `claim_overlay()` from a JSX
+    // `<View overlay>` mount and `release_overlay()` from its unmount.
+    // The ComboBox path remains untouched and has its own state.
+    static View* active_overlay_;
+    void claim_overlay() { active_overlay_ = this; }
+    /// Clear the global overlay if (and only if) `this` currently holds it.
+    /// Idempotent — safe to call on unmount even if claim never happened.
+    void release_overlay() {
+        if (active_overlay_ == this) active_overlay_ = nullptr;
+    }
+    /// Bounds-test in window (root) coordinates. Walks the parent chain
+    /// to compute absolute origin and adds local_bounds(). Mirrors the
+    /// arithmetic the mac mouseDown path uses for the ComboBox dropdown.
+    bool overlay_contains(Point window_pt) const;
 
     /// Global click callback (fires on any view click with widget id). Set on root.
     std::function<void(const std::string& id, uint16_t modifiers)> on_global_click;
--- a/core/view/platform/mac/window_host_mac.mm
+++ b/core/view/platform/mac/window_host_mac.mm
@@ -373,6 +373,55 @@ static pulp::view::KeyCode keyCodeFromNS(unsigned short code) {
             }
         }
 
+        // pulp #1148 — generalized overlay-click routing for React popovers.
+        // Any View that called `claim_overlay()` (e.g. via @pulp/react's
+        // `<View overlay>` JSX prop) is checked AFTER the ComboBox path
+        // (which stays exact-as-was per regression test in
+        // test_combo_dropdown.cpp [issue-overlay]) and BEFORE the regular
+        // tree hit_test. If the click falls inside the overlay's window
+        // rect we route it directly there so absolutely-positioned popover
+        // children get the click instead of whatever sibling/ancestor view
+        // happens to occupy that pixel.
+        if (auto* overlay = pulp::view::View::active_overlay_) {
+            if (view_is_in_tree(overlay, self.rootView) &&
+                overlay->overlay_contains({pt.x, pt.y})) {
+                // Hit-test inside the overlay subtree so nested buttons /
+                // labels still receive the click. Falls back to the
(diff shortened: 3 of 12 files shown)


danielraffel:
Force-pushed `feature/overlay-click-routing-1148` (per coordination protocol — pulp-side owns version bumping):

1. Rebased onto v0.70.0 main
2. Added `chore: bump versions` to bump SDK 0.70.0 → 0.71.0 (your #1148 fix is user-facing feat that needs a release tag)

Note: #1060 has the same target (0.71.0) so whichever merges first wins; the loser will need to rebase + re-bump to 0.72.0. CI re-running now.

The diff-coverage failure is the same `pulp_import_design.cpp` exclusion gap we hit on #1060 (#1290) — but your changes are in `core/view/` not `tools/import-design/`, so this should clear once CI re-evaluates against fresh main.

danielraffel:
Re-pushed `feature/overlay-click-routing-1148` (per coordination protocol):

- Rebased onto fresh main (auto-dropped the previous bump since #1060 already shipped 0.71.0)
- Bumped 0.71.0 → 0.72.0
- Added 7× `Compat-Update: skip prefix=<…>` trailers on the bump commit. The compat-sync gate fired on every prefix because window_host changes touch a paths-mapped surface, but this PR doesn't add new CSS props / RN style props / Yoga layout / React intrinsics / HTML semantics / canvas2d / import formats — it routes existing clicks correctly.

CI re-running. The diff-coverage failure (separate from version+skill) was on platform window_host code that's hard to unit-test without spinning up a real window — likely needs a `--exclude=window_host_*.{mm,cpp}` similar to the pulp_import_design.cpp pattern. Will help if it persists after this push.

PR #1327, opened 2026-05-03
fix(view, mac): NSWindow resize triggers root View relayout (closes pulp #1321)

### LINKED ISSUE #1321: view: window resize doesn't reflow content — Yoga layout cuts off, doesn't re-flow on size change
## Summary

When the Spectr standalone window is resized at runtime (drag the resize corner, or programmatically `set size of window 1`), **the content does NOT reflow**. Visible content gets clipped at the new window edges. The Yoga layout tree appears to be using the original window size for layout and isn't re-laying out on resize events.

## Repro

1. Launch Spectr standalone (built against pulp v0.72.1)
2. Default size (1320x892): renders correctly
3. Drag the resize corner OR `osascript -e 'tell application "System Events" to tell process "Spectr" to set size of window 1 to {800, 500}'`
4. Window shrinks
5. Content does NOT reflow — toolbar/spectrum/etc. retain their original positions and get clipped at the new window edge
6. Resize back to 1320x892 — content stays in original layout
7. Compare WebView Spectr — resizing reflows everything (browser layout is reactive to viewport size)

## Hypothesis

Pulp's window-host platform (e.g. `core/view/platform/mac/window_host_mac.mm`) probably doesn't:

1. Listen for `NSWindowDidResizeNotification` (or equivalent on x11/win32/iOS)
2. On resize: update the root View's bounds AND trigger a Yoga layout recompute on the descendants
3. Repaint after re-layout

Or possibly: the resize signal does fire, but Yoga's `setStyleWidth/setStyleHeight` on the root isn't being called with the new dimensions — so Yoga recomputes against the old constraints.

## Impact

P1 — every Pulp standalone or DAW-hosted plugin with a resizable window hits this. WebView consumers don't because Chromium does layout reflow for free.

Spectr in particular runs in standalone mode for testing; once it ships as a plugin in a DAW, the host can resize the editor view and content WILL get clipped without this fix.

## Cross-refs

- Spectr branch: `feature/native-react-editor` @ `c71ae3e`
- SDK pin: v0.72.1
- Related: #998 (App-root layout regression — Spectr's `346eb64` workaround uses explicit width/height because automatic propagation never worked)
- Potentially related to #1147 popover-render in that nested flex doesn't reflow correctly

## Suggested fix shape

In `window_host_mac.mm`:
1. Implement `windowDidResize:` NSWindow delegate (or wire NotificationCenter observer)
2. On resize, set the root View's bounds to the new content rect
3. Call `View::layout()` (the existing Yoga relayout entry point) to recompute descendants
4. `setNeedsDisplay:YES` to trigger paint

Mirror in iOS host (`window_host_ios.mm`) for `viewDidLayoutSubviews` — touch-driven resize on rotation.

danielraffel:
## Summary

Closes #1321. Spectr's standalone window resize now reflows content correctly — top toolbar (LIVE/PRECISION/IIR/FFT/HYBRID/bands/zoom) + bottom toolbar (CLEAR/SCULPT/PEAK/PRESETS/SNAPSHOT/A·B/spread) reflow proportionally on resize, nothing clips, round-trip is bit-stable.

## Root cause (three-way)

1. **PulpView (CPU)** had no `autoresizingMask` — contentView's `bounds.size` stayed at the original frame even though `drawRect:` was reading it
2. **No `setFrameSize:` override** pushed new bounds into the `rootView`
3. **`windowDidResize:`** only forwarded to a user `resize_callback`, never to the root view itself
4. GPU path deferred relayout to the next vsync via `needs_repaint_`, leaving hit-testing stale during drag-resizes

## Fix

- Set `autoresizingMask` in `PulpView::initWithFrame:`
- New `PulpView::setFrameSize:` calls `rootView->set_bounds()` + `layout_children()` + `setNeedsDisplay:YES`
- Both `PulpWindowDelegate.onResize` blocks (CPU + GPU) now relayout root before forwarding to the user callback
- GPU `handle_resize()` runs the relayout synchronously instead of waiting for vsync

## Test plan

- [x] Catch2 `[issue-1321]` test in `test/test_view.cpp` — model-path resize reflows children, fixed-size sibling preserved next to flex child
- [x] Full `pulp-test-view`: 49 cases / 236 assertions, all green
- [x] **End-to-end Spectr standalone validation**: programmatic resize 1320×892 → 800×500 → 1320×892. Toolbars reflow, content scales, nothing clips, round-trip bit-stable. Screenshots in `pulp-planning/screenshots/issue-1321/` linked from the issue comment.
- [ ] CI green on macos / linux / windows
- [ ] iOS host mirror (`viewDidLayoutSubviews`) — flagged as follow-up; left as separate slice since there's no iOS Spectr build to validate against

## Files

- `core/view/platform/mac/window_host_mac.mm` (+42 lines)
- `test/test_view.cpp` (+97 lines, +1 include)

## Cross-refs

- Issue: #1321
- Closes the "window resize cuts off content" symptom user reported on Spectr v0.72.1
- Filed during the 2026-05-03 audit pass (`spectr/planning/audit-2026-05-03-webview-vs-native-v0.69.1.md`)

## Validation evidence

Local Spectr build at `/tmp/spectr-1321/build/Spectr.app` against `~/.pulp/sdk/0.72.2-rc1`. Resize before/after screenshots committed to `pulp-planning/screenshots/issue-1321/` (private submodule). Issue comment 4366472274 has direct links.

### DIFF
--- a/core/view/platform/mac/window_host_mac.mm
+++ b/core/view/platform/mac/window_host_mac.mm
@@ -209,6 +209,19 @@ static std::vector<uint8_t> capture_window_screencapture_png(NSWindow* window) {
     pulp::view::View* _focusedView;
 }
 
+- (instancetype)initWithFrame:(NSRect)frame {
+    self = [super initWithFrame:frame];
+    if (self) {
+        // pulp #1321: ensure the content view tracks the window's content rect
+        // so AppKit resizes our frame when the user drags the window edge.
+        // Without this, the view's bounds stay at the original frame and Yoga
+        // never reflows on resize.
+        self.autoresizesSubviews = YES;
+        self.autoresizingMask = NSViewWidthSizable | NSViewHeightSizable;
+    }
+    return self;
+}
+
 - (BOOL)isFlipped { return NO; }
 - (BOOL)acceptsFirstResponder { return YES; }
 - (BOOL)acceptsFirstMouse:(NSEvent*)e { (void)e; return YES; }
@@ -1024,6 +1037,20 @@ static pulp::view::KeyCode keyCodeFromNS(unsigned short code) {
     }
 }
 
+- (void)setFrameSize:(NSSize)newSize {
+    [super setFrameSize:newSize];
+    // pulp #1321: AppKit resizes us during a window resize. Push the new
+    // bounds into the root View immediately and relayout so hit testing,
+    // tracking-area updates, and the next paint all see the new geometry.
+    if (self.rootView) {
+        self.rootView->set_bounds({0, 0,
+            static_cast<float>(newSize.width),
+            static_cast<float>(newSize.height)});
+        self.rootView->layout_children();
+    }
+    [self setNeedsDisplay:YES];
+}
+
 - (void)dealloc {
     [self.animationTimer invalidate];
 }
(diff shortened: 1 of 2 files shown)


danielraffel:
Amended tip commit with Version-Bump: skip + Skill-Update: skip trailers per coordination protocol — pulp-side owns version-bump hygiene; both gates were the only remaining blockers. Force-pushed; CI re-running.

PR #1328, opened 2026-05-03
fix(view): overlay_contains walks overflow:visible children (closes pulp #1320)

### LINKED ISSUE #1320: view: bands picker hit-test is asymmetric — 32/40 don't dispatch but 48/56/64 do (Spectr v0.72.1)
## Summary

In Spectr at v0.72.1 (with #1297 overlay-click routing landed), the bands-picker dropdown opens on tap but inner-click dispatch is **asymmetric**:

- **Right half (48, 56, 64) DOES dispatch** — bands change as expected
- **Left half (32, 40) does NOT dispatch** — clicks no-op silently

This was originally noted in the v0.68.0 audit (`spectr/planning/audit-2026-04-30-native-vs-webview.md`) as the one-directional `overflow:visible` hit-test bounds bug. v0.72.1's #1297 generalized overlay-click routing was expected to fix this, but the asymmetry persists.

## Repro

1. Build Spectr standalone against pulp v0.72.1 SDK + latest pulp-react bundle
2. Run `./build/Spectr.app/Contents/MacOS/Spectr`
3. Click the bands indicator (top-right "32 bands ▾" or similar)
4. Bands picker dropdown opens with 5 cells: 32 / 40 / 48 / 56 / 64
5. Click "32" → no change
6. Click "40" → no change
7. Click "48" → bands changes to 48 ✓
8. Click "56" → bands changes to 56 ✓
9. Click "64" → bands changes to 64 ✓

## Hypothesis

The bands picker is a `<View overlay>` (post-#1297) but the picker's `local_bounds()` doesn't include the leftmost cells — they paint outside the View's hit-test rect. When a click lands at the 32/40 column, `View::active_overlay_->overlay_contains(point)` returns false and the click falls through to the tree hit_test underneath.

Verify by:
- Logging `View::active_overlay_->overlay_contains(pt)` result in `mouseDown` for clicks in the 32 vs 64 cell positions
- Checking the bands picker's bounds + `overflow_` flag at runtime

## Cross-refs

- Originally documented in v0.68.0 audit + #1148 design pointer
- #1297 expected to fix; #1316 was Codex P1 follow-up
- May indicate `overlay_contains` needs to include children rendered via `overflow:visible` outside parent bounds

## Spectr-side

- `feature/native-react-editor` @ `c71ae3e`
- editor.js bundle: 451KB (deduped per #1295)
- 5/5 direct-exec stable

danielraffel:
## Summary

Closes #1320. Generalizes `View::overlay_contains` to include the painted bounding box of `overflow:visible` descendants, matching CSS semantics. Without this expansion, clicks on overflow children of an overlay View fell through to tree hit_test.

## Tests

3 new Catch2 cases tagged `[issue-1320]`:
- includes overflow:visible child painted outside parent bounds — pass
- stops walking at overflow:hidden child — pass  
- own overflow:hidden disables expansion — pass

`pulp-test-overlay-routing`: 31/31 assertions, 10/10 cases pass.

## Important caveat re: the bands-picker bug

This fix lands the framework primitive correctly, but Spectr's `editor.js` bundle doesn't currently use `<View overlay>` for the bands picker — it uses bare `<div style="position:absolute">`. So this fix is unreachable for the specific symptom on Spectr today.

The bands picker bug (32/40 fail, 48/56/64 work) actually triggers in `View::hit_test()` at `core/view/src/view.cpp:475-480` where the existing `overflow:visible` extension is asymmetric — only +500px DOWNWARD. The bands picker extends LEFTWARD beyond its narrow trigger button, so the parent's hit_test returns nullptr before the click reaches the absolute-positioned dropdown child.

Two possible follow-ups:
1. Make `View::hit_test`'s overflow:visible extension symmetric (left/right/up, not just down) — separate slice
2. Spectr-side bundle change to opt into `overlay={true}` on the bands-picker wrapper

This PR addresses (1) the framework primitive; (2) is bundle-side; the symmetric hit_test extension is the next slice if we want overflow-visible to "just work" without explicit overlay opt-in.

## Files

- `core/view/src/view.cpp` (+69 lines)
- `test/test_overlay_routing.cpp` (+113 lines, 3 cases)

## Cross-refs

- Issue: #1320
- Related: #1148 / #1297 (overlay click routing), #1316 (Codex P1 follow-up)
- Spectr-side: bands picker still asymmetric until either #1320b (symmetric extension) or bundle opt-in

### DIFF
--- a/core/view/src/view.cpp
+++ b/core/view/src/view.cpp
@@ -529,6 +529,44 @@ bool View::call_inspector_mouse_hook(const MouseEvent& e) {
 // pulp #1148 — generalized overlay-click routing.
 View* View::active_overlay_ = nullptr;
 
+namespace {
+
+// pulp #1320 — recursively expand a child's painted-bounds contribution
+// up through any `overflow:visible` descendants. Returns the bounding
+// rect (in window coords) of `v` and every transitive descendant whose
+// chain back to `v` is entirely overflow:visible. A descendant inside
+// an `overflow:hidden` ancestor is clipped, so it stops contributing.
+//
+// `parent_abs_x` / `parent_abs_y` are the absolute window-coord origin
+// of `v->parent()`. The function consumes those, applies `v`'s own
+// `bounds().x/y`, and recurses.
+void accumulate_overflow_extent(const View* v,
+                                float parent_abs_x,
+                                float parent_abs_y,
+                                float& min_x,
+                                float& min_y,
+                                float& max_x,
+                                float& max_y) {
+    if (!v) return;
+    const float abs_x = parent_abs_x + v->bounds().x;
+    const float abs_y = parent_abs_y + v->bounds().y;
+    const auto lb = v->local_bounds();
+    if (abs_x < min_x) min_x = abs_x;
+    if (abs_y < min_y) min_y = abs_y;
+    if (abs_x + lb.width > max_x) max_x = abs_x + lb.width;
+    if (abs_y + lb.height > max_y) max_y = abs_y + lb.height;
+    // Only recurse through children whose own overflow is visible —
+    // that's the CSS rule. An `overflow:hidden` child clips its own
+    // descendants, so they don't contribute painted pixels above us.
+    if (v->overflow() != View::Overflow::visible) return;
+    for (size_t i = 0; i < v->child_count(); ++i) {
+        accumulate_overflow_extent(v->child_at(i), abs_x, abs_y,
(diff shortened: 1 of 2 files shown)


danielraffel:
Amended tip commit with Version-Bump: skip + Skill-Update: skip trailers per coordination protocol — pulp-side owns version-bump hygiene; both gates were the only remaining blockers. Force-pushed; CI re-running.

PR #1329, opened 2026-05-03
fix(canvas/mac): CoreGraphicsCanvas implements path / gradient / transform (closes pulp #1322)

### LINKED ISSUE #1322: view: empty FilterBank canvas at v0.70.0+ — analyzer feed regression after React-instance dedup (#1295)
## Summary

Spectr's FilterBank canvas (the central spectrum visualization) **rendered correctly at v0.69.1** — full color gradient (blue → green → yellow → red) over a 32-band-shaped curve. **At v0.70.0+ (the release that ships #1295's React-instance dedup) the canvas renders empty white** with no spectrum data.

The `editor.js` bundle changed in size at the SDK upgrade (1.25 MB → 451 KB) because the duplicate React was removed by #1295's `--external:react/...` change to @pulp/react's build script. The canvas regression coincides exactly with that bundle change.

## Repro

1. Build Spectr against v0.70.0 (or v0.71.0 / v0.72.0 / v0.72.1 — same behavior on all)
2. With the rebuilt `editor.js` (deduped React, 451KB)
3. Launch direct-exec
4. Top + bottom toolbar render correctly; tabs work; popover panels open
5. **Center FilterBank area: empty white canvas — no spectrum, no band shapes, no gradient**

Comparison: at v0.69.1 with the OLD `editor.js` (pre-dedup, 1.25MB, two React instances), the spectrum DID render. Visible in `spectr/planning/screenshots/native-editor-v0.69.1-idle.png`.

## Hypothesis

The bundle change is the only delta. `spectr-editor-extracted.js` (the Claude Design extracted React app) may rely on a React-internal access pattern that worked when the user-bundle had its own React, and broke when @pulp/react's reconciler now operates on the same React instance.

Candidates for the failing pattern:
1. **`useEffect` timing for analyzer subscription** — the FilterBank component subscribes to Spectr's analyzer feed on mount via `useEffect`. With deduped React, effect-flush ordering may differ enough that the subscription happens before `setSpectrumData` is callable, OR the subscription's cleanup runs before the first frame.
2. **Ref forwarding through the deduped reconciler** — the canvas widget gets its `id` via `useRef` then passes it to native `setSpectrumData`. With deduped React, `useRef` semantics under the new reconciler may differ.
3. **Bridge globals racing with reconciler init** — the analyzer feed depends on `globalThis.setSpectrumData` being installed by `spectr_editor_assets` before the JS engine evaluates `editor.js`. The deduped bundle may evaluate faster and miss the bridge install.

## Investigation seam

- Add `console.log` (or `__spectrLog`) at the FilterBank component's `useEffect` mount + cleanup to verify subscription fires
- Add bridge-side logging when `setSpectrumData` is first called
- Compare timing between v0.69.1 (working) and v0.72.1 (broken)

## Spectr-side fix may be sufficient

This is likely a Spectr-side bundle issue (the analyzer wiring in `host-shims.ts` / `dom-adapter.tsx` / `spectr-editor-extracted.js`), not pulp framework. But since the symptom only appeared after pulp shipped #1295, there's a chance pulp's `setSpectrumData` semantics changed subtly. Filing here so it gets eyes from both sides.

## Cross-refs

- #1295 (the React dedup that ships in v0.70.0)
- #964 (original FilterBank canvas surface — was closed at v0.68.1+; this is a NEW regression)
- v0.69.1 working baseline: `spectr/planning/screenshots/native-editor-v0.69.1-idle.png`
- v0.72.1 broken: `spectr/planning/screenshots/native-editor-v0.72.1-idle.png`

## Spectr branch

- `feature/native-react-editor` @ `c71ae3e`
- editor.js: 451KB (deduped)
- 5/5 direct-exec stable; only the analyzer canvas is empty

danielraffel:
## Summary

Closes #1322. Spectr's FilterBank canvas regressed from "rendering full color gradient" (v0.69.1) to "empty white" (v0.70.0+). Root cause: **`CoreGraphicsCanvas` (CPU backend used by `run_with_editor(use_gpu=false)`) inherited no-op base-class defaults for every Canvas2D-style path / gradient / transform method.**

## Diagnosis

`core/canvas/include/pulp/canvas/cg_canvas.hpp` declared only the shape API + `concat_transform`. The base class `Canvas` declared as `{}` no-ops:

- `begin_path` / `move_to` / `line_to` / `quad_to` / `cubic_to` / `close_path`
- `fill_current_path` / `stroke_current_path`
- `set_fill_gradient_linear` / `set_fill_gradient_radial` (fell back to single solid color)
- `set_transform` / `clip`

Spectr's stderr trace shows ~1800 `canvasBeginPath` / `canvasFillPath` / `canvasMoveTo` / `canvasLineTo` / `canvasSetLinearGradient` / `canvasSetTransform` commands per frame — **all** were dispatched correctly by the bridge and **all** silently dropped at paint time. Only `canvasClearRect` + `fill_rect` got through.

`SkiaCanvas` (GPU) implements all of these — explains why GPU mode rendered the spectrum and CPU mode didn't.

## Fix

Implements the missing surface in `cg_canvas.hpp/.mm`:
- Path building via `CGMutablePathRef`
- Gradients via `CGContextDrawLinearGradient` / `CGContextDrawRadialGradient` clipped to the path / rect
- `set_transform` + `capture_paint_baseline_transform` with proper composition onto the captured baseline (mirrors SkiaCanvas semantics)

## Test plan

4 new Catch2 cases under tag `[issue-1322]` in `test/test_canvas.cpp`:

- Diamond polygon fill — pass
- Bezier curve fill — pass
- Polyline stroke — pass
- Linear gradient with two stops — pass

All 7 existing CG tests still green. Full `test_canvas`: 36 cases / 1289 assertions ✓. `test_canvas_widget`: 17 cases ✓.

## Files changed

- `core/canvas/include/pulp/canvas/cg_canvas.hpp` — +declarations
- `core/canvas/platform/mac/cg_canvas.mm` — +implementations
- `test/test_canvas.cpp` — +4 cases

## End-to-end Spectr validation

Branch SHA `184c478b`. Spectr currently pinned to `~/.pulp/sdk/0.72.1`. After this PR merges + auto-release tags v0.72.2 / v0.73.0, will:
1. `pulp sdk install --version <new>`
2. Rebuild Spectr standalone
3. Launch direct exec — spectrum should now render in CPU mode

Capture before/after will be added as a comment on this PR + the issue.

## Cross-refs

- Issue: #1322
- Original v0.69.1 vs v0.70.0+ regression timeline tracked in `spectr/planning/audit-2026-05-03-webview-vs-native-v0.69.1.md`
- Adjacent: SkiaCanvas already had this surface; this PR brings CPU parity

## Closing context (per "document fixes for closure" protocol)

When this merges and #1322 closes:
- Root cause: CG canvas no-op inheritance
- Fix: implement path / gradient / transform in `cg_canvas.mm`
- Verify: launch Spectr standalone with `gpu=false` (default) → FilterBank spectrum renders
- Caveat: GPU mode was always fine; this fix is CPU-mode-specific

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.72.2
+    VERSION 0.72.3
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/canvas/include/pulp/canvas/cg_canvas.hpp
+++ b/core/canvas/include/pulp/canvas/cg_canvas.hpp
@@ -2,10 +2,13 @@
 
 #include <pulp/canvas/canvas.hpp>
 
+#include <vector>
+
 #ifdef __APPLE__
 
-// Forward declare CGContext
+// Forward declare CG types
 typedef struct CGContext* CGContextRef;
+typedef struct CGPath* CGMutablePathRef;
 
 namespace pulp::canvas {
 
@@ -23,13 +26,25 @@ public:
     void rotate(float radians) override;
     void concat_transform(float a, float b, float c,
                           float d, float e, float f) override;
+    void set_transform(float a, float b, float c,
+                       float d, float e, float f) override;
+    void capture_paint_baseline_transform() override;
     void clip_rect(float x, float y, float w, float h) override;
+    void clip() override;
 
     void set_fill_color(Color c) override;
     void set_stroke_color(Color c) override;
     void set_line_width(float w) override;
     void set_line_cap(LineCap cap) override;
     void set_line_join(LineJoin join) override;
+
+    void set_fill_gradient_linear(float x0, float y0, float x1, float y1,
+                                   const Color* colors, const float* positions,
+                                   int count) override;
+    void set_fill_gradient_radial(float cx, float cy, float radius,
+                                   const Color* colors, const float* positions,
+                                   int count) override;
+    void clear_fill_gradient() override;
 
     void fill_rect(float x, float y, float w, float h) override;
     void clear_rect(float x, float y, float w, float h) override;
@@ -43,6 +58,18 @@ public:
     void stroke_line(float x0, float y0, float x1, float y1) override;
     void stroke_path(const Point2D* points, size_t count) override;
     void fill_path(const Point2D* points, size_t count) override;
+
+    // Canvas2D-style path building (pulp #1322).
+    void begin_path() override;
+    void move_to(float x, float y) override;
+    void line_to(float x, float y) override;
+    void quad_to(float cpx, float cpy, float x, float y) override;
+    void cubic_to(float cp1x, float cp1y, float cp2x, float cp2y,
+                  float x, float y) override;
+    void close_path() override;
+    void fill_current_path() override;
+    void stroke_current_path() override;
+
     void set_opacity(float alpha) override;
     void save_layer(float x, float y, float w, float h,
                     float opacity, float blur_radius) override;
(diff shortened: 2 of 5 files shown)


danielraffel:
## ✅ End-to-end Spectr validation passed (with caveat)

Built against `pulp v0.72.1` (NOT yet this PR) + the Spectr-side `canvas2d-shim.ts` setFlex sizing fix at `spectr@ce47893`. **Spectrum renders correctly** in CPU mode.

This is interesting because the no-op bug this PR fixes IS real — but the visible Spectr symptom resolves without it, because Spectr's canvas drawing leans on `fill_rect` + `clearRect` which the CG backend already supports. The path / gradient / transform implementations this PR adds are needed for any plugin whose canvas content uses richer primitives.

**Net**: this PR is still the right fix and should land. Spectr just happens to not exercise the broken paths heavily enough to have been blocked end-to-end.

When PR #1329 merges + ships in a release, recommend closing #1322 referencing both `spectr@ce47893` (Spectr-side sizing fix) AND this PR (CG backend completeness).

danielraffel:
Added `chore: bump versions` (0.72.1 → 0.73.0) to satisfy version-bump gate (per coordination protocol — pulp-side owns version-bump hygiene). This is a real user-facing canvas fix that should ship in v0.73.0.

The diff-coverage failure is separate — `core/canvas/platform/mac/cg_canvas.mm` is platform-specific and hard to unit-test without spinning up a real CG context. If it doesn't clear with the rebuild, we may need a `diff_cover_excludes` entry similar to what landed for `pulp_import_design.cpp` in #1290.

Issue #1548, opened 2026-05-06
rn: wire textShadow cluster (Color/Offset/Radius) (refs #1434, Phase A3)


danielraffel:
**Coverage gap:** 3 rn NOT-IMPLs:
- `rn/textShadowColor`
- `rn/textShadowOffset` ({width, height})
- `rn/textShadowRadius`

**Status:** rn projected ~89% after #1547 lands. +3 entries → **~92%**.

**Scope:**
- New bridge setter: `viewSetTextShadow(color int_val, offsetX float, offsetY float, radius float)` — packed args.
- Skia paint path: `SkPaint::setShadowMaskFilter` (already used elsewhere in canvas2d shadow #1446).
- `prop-applier.ts`: route 3 props.
- Catalog flip: 3× missing → supported.

**Tests:** SkPaint mask filter asserted in headless render fixture.

Refs umbrella: #1434.

Source: `/tmp/rn-api-audit.md`.

danielraffel:
## Hourly triage (automated)

**Scope:** 3 RN text-shadow props are NOT-IMPL (`textShadowColor`, `textShadowOffset` `{width, height}`, `textShadowRadius`) — wiring them to Skia's `SkPaint::setShadowMaskFilter` (already used by canvas2d shadow work in #1446) would add +3 rn entries toward 92%.

**Complexity:** small-medium — new packed bridge setter `viewSetTextShadow(color, offsetX, offsetY, radius)` in `widget_bridge.cpp`; Skia paint path via `SkBlurMaskFilter`; 3 `prop-applier.ts` prop routes; catalog flip.

**Implementation plan:**
- `widget_bridge.cpp`: register `viewSetTextShadow(color_int, offsetX_float, offsetY_float, radius_float)` as a packed setter (avoids 3 separate bridge round-trips).
- Skia paint path: call `SkPaint::setShadowMaskFilter(SkMaskFilter::MakeBlur(kNormal_SkBlurStyle, radius))` with the color and offset applied via `SkPaint::setColor` + layer translation.
- `prop-applier.ts`: accumulate `textShadowColor`, `textShadowOffset`, `textShadowRadius` into one `viewSetTextShadow` call when all 3 are present.
- Catalog flip: 3× `missing` → `supported`; add `SkPaint` mask-filter assertion in a headless render fixture.

**Priority signal:** P2 — additive RN API coverage; text shadows are common in typography-rich audio UI designs; shares Skia infrastructure already wired for canvas2d shadows.

[...]

Issue #1434, opened 2026-05-05
META: Framework-import-readiness — render Figma/Pencil/v0/Claude Design/RN exports with near-zero tweaking


danielraffel:
## Goal

Pulp's framework value = **imports from common prototyping/design tools render through pulp's Skia/Dawn pipeline with near-zero developer tweaking, matching webview fidelity**.

Import sources of record (the test surface):

- **Figma** exports (export-as-code or Figma plugins)
- **Pencil.dev** (MCP)
- **Open Design** (`/Users/danielraffel/Code/open-design`)
- **v0.dev** (MCP / file exports)
- **Claude Design** (react) HTML
- **React Native** file exports
- **Generic HTML/CSS/React** designs that render well in a browser

Spectr is one consumer test bed among N, useful as a canary; not the validation target.

## Quality bars (joint)

1. **Property coverage**: per-surface **PASS+DIVERGE ≥ 95%** measured by the property-coverage harness (pulp #1391 / #1392 / #1393 / #1394). DIVERGE counts because it means the property is wired in some form.
2. **OOS audit-clean**: every OOS entry justifiable as "this can't reasonably exist on pulp's native runtime" (e.g. paged-media, browser DOM-only events). Anything import sources commonly emit cannot be OOS.
3. **Visual fidelity**: paint-correctness harness sibling (pulp #1409) wired into CI; per-fixture pixel-diff against webview reference within tolerance.
4. **Fixture library**: representative test imports per source (Figma export, Pencil document, v0 output, Claude Design HTML, RN export, generic HTML/CSS sample) committed under `test/fixtures/import-design/`. Imported through pulp's render path; diffed against webview render.

## Per-surface coverage scorecard (2026-05-05, post-#1422 main)

| Surface | Total | PASS | DIVERGE | NOT-IMPL | OOS | drift | PASS% | PASS+DIVERGE% | Target | Gap |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| yoga | 53 | 8 | 28 | 17 | 0 | 22 | 15.1% | **67.9%** | 95% | −27 pp |
| css | 195 | 23 | 85 | 47 | 31 | 72 | 11.8% | **55.4%** | 90% | −35 pp |
| canvas2d | 63 | 34 | 18 | 11 | 0 | 5 | 54.0% | **82.5%** | 95% | −13 pp |
[...]

PR #1569, opened 2026-05-06
fix(rn): textShadow cluster — Color/Offset/Radius paint-time glyph shadow (closes pulp #1548)

danielraffel:
## Summary

Closes pulp #1548 — flips three RN style props from `missing` → `supported` in the rn surface:

- `rn/textShadowColor`
- `rn/textShadowOffset` (RN `{ width, height }` object)
- `rn/textShadowRadius`

Paint-time only — applied to glyphs via `Label::paint`, no Yoga layout impact. Default `alpha=0` keeps the path a no-op until JS opts in.

## Changes

**View slots (`core/view/include/pulp/view/view.hpp`)** — four new fields: `text_shadow_color_` (default fully transparent), `text_shadow_offset_x_/y_`, `text_shadow_radius_`, plus get/set accessors.

**Bridge (`core/view/src/widget_bridge.cpp`)** — three new fns: `setTextShadowColor`, `setTextShadowOffset(id, dx, dy)`, `setTextShadowRadius`. Each writes ONE slot in isolation so a JSX prop diff that touches one does not clobber the others.

**Label paint (`core/view/src/widgets.cpp`)** — copies the trio into the canvas shadow API (`set_shadow_color` / `set_shadow_blur` / `set_shadow_offset_x/y`) just before `fill_text` and clears them right after, so glyph shadows don't leak into decoration strokes (underline / line-through) or downstream siblings. On the Skia backend `apply_shadow_filter` installs an `SkBlurMaskFilter` via `SkImageFilters::DropShadow` — same code path canvas2d shadows already use (#1446).

**@pulp/react** — `prop-applier.ts` adds three cases. `textShadowColor` and `textShadowRadius` pass verbatim. `textShadowOffset` is the RN `{ width, height }` object — split into `(dx, dy)` before bridge dispatch (defensive coercion to 0 when an axis is missing). `types.ts` + `bridge.ts` carry the matching declarations + mock-bridge entries.

**Catalog (`compat.json`)** — three rn entries flipped `missing` → `supported` with full `mapsTo`, `supportedValues`, `tests`, and `notes` payloads.

**Docs (`docs/reference/compat/rn.md`)** — Recently changed bullet leading the section.

## Tests

- `test/test_widget_bridge.cpp` — 4 `[issue-1548]` Catch2 cases (56 assertions): bridge round-trip + per-attribute isolation; `Label::paint` emits `set_shadow_*` before AND after `fill_text` (set + clear); default state emits zero shadow commands; alpha=0 with non-zero offset/radius still short-circuits.
- `packages/pulp-react/test/prop-applier-text-shadow.test.ts` — 6 vitest cases: per-prop verbatim forwarding, RN `{ width, height }` split, partial-offset coercion to 0, three-prop coexistence, no cross-talk with outline / boxShadow clusters.

## Test plan

- [x] `./build/test/pulp-test-widget-bridge "[issue-1548]"` — 4 cases / 56 assertions pass
- [x] `./build/test/pulp-test-widget-bridge` — full suite 1190 assertions / 175 cases pass
- [x] `./build/test/pulp-test-widgets` — 260 assertions / 64 cases pass
- [x] `pnpm test` (packages/pulp-react) — 238 tests / 26 files pass

Refs umbrella: pulp #1434 (Phase A3 Bundle 2).


danielraffel:
Closing as obsolete — the RN textShadow cluster work was already wired to main via #1661 (`Color`-typed `View::set_text_shadow_color/offset/radius` in `core/view/include/pulp/view/view.hpp` line 920+). This PR proposed a parallel `std::string`-typed implementation that conflicts at the type level with what shipped. The catalog drift this PR was trying to fix was also resolved in #1661.

If there are textShadow gaps still on main (e.g. RN `textShadowOffset` array form vs flat dx/dy), please file a fresh issue with the specific gap — the existing `Color`-based slots would be the integration point.

Issue #1708, opened 2026-05-10
crash(mac): use-after-free on _focusedView when focused widget is destroyed


danielraffel:
## Summary

`PulpView` (`core/view/platform/mac/window_host_mac.mm`) holds the focused widget as a raw `pulp::view::View*` pointer in `_focusedView`. When that widget is destroyed by a tree update (e.g., a React reconciler unmounts the focused TextEditor or its ancestor), `_focusedView` becomes dangling. The next key event triggers `-[PulpView focusedTextEditor]` which calls `dynamic_cast<TextEditor*>(_focusedView)` on the dangling pointer — segfault.

## Crash signature

```
0   libc++abi.dylib      __cxxabiv1::dyn_cast_get_derived_info
1   libc++abi.dylib      __dynamic_cast
2   AppKit               -[NSTextInputContext hasMarkedTextWithCompletionHandler:]
…
30  AppKit               -[NSView interpretKeyEvents:]
31  Spectr               -[PulpView keyDown:] + 1252
```

`Exception Type: EXC_BAD_ACCESS  KERN_INVALID_ADDRESS at 0xc90028b9c03dc540 (possible pointer authentication failure)` — classic dangling-pointer signature.

## Repro

Spectr's native bridge editor. Mount any view, navigate around (e.g., open Settings, close it, draw bands), then press any key on the keyboard. Crash happens consistently when the focus tracking pointer outlives the widget. Closing or unmounting the focused view doesn't clear `_focusedView`, and a later key event reaches the dangling pointer.

## Suggested fix

Two options:

1. **Weak observation:** Make `View` track focus state internally with a `bool focused_` flag, and have `PulpView` look up the focused widget by walking the tree each time. Slow but eliminates dangling pointers.

2. **Notify-on-destroy hook:** When a `View` is destroyed and it's currently focused, walk up to its `WindowHost` and clear `_focusedView`. The View destructor would call something like `if (auto* host = window_host()) host->clear_focus_if(this);`.

3. **Defensive nil-check at use sites:** Track focus generation. When `_focusedView` is set, also store the View's `id` (string) and look it up via `widgets_[id]` map at use time, returning `nullptr` if it's gone. Bridge layer already has this map.

Currently the failure mode (segfault on every keypress after a relevant unmount) makes the framework unusable for any app that mutates its view tree based on user interaction.

## Reference

[...]

PR #1726, opened 2026-05-10
fix(view): clear input-focus slot on widget destruction (#1708)

danielraffel:
Use-after-free in -[PulpView keyDown:] -> -[PulpView focusedTextEditor]
when the focused widget is destroyed by a tree update (e.g., a React
unmount of an open Settings modal). PulpView held a raw View* in
_focusedView; the next keypress dynamic_cast<TextEditor*> on the freed
memory segfaulted inside libc++abi.

Mirror the pulp #1148 active_overlay_ pattern: add a static
View::focused_input_ slot that the destructor auto-clears, and route
the macOS window host's keyDown / text-input dispatch through it
instead of its own raw ivar.

- View::focused_input_ static, claim_input_focus() / release_input_focus()
- ~View() clears the slot if this == focused_input_
- PulpView calls claim/release on focus changes and on
  clearInteractionState; reads focused_input_ for hasMarkedText,
  selectedRange, setMarkedText, unmarkText, attributedSubstring,
  insertText, focusedTextEditor

Verified end-to-end via Spectr's native-react editor: Settings can
now be opened, dismissed via Esc/click-outside, and typed into without
crashing. Repro path (close modal -> press any key -> segfault) now
exits cleanly through the auto-cleared slot.

5 regression tests in test/test_focused_input.cpp cover: default
nullptr, claim, mutual release, swap-on-destruct (the actual UAF
guard), and sibling destruct doesn't clear holder.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.82.1
+    VERSION 0.82.2
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/view/include/pulp/view/view.hpp
+++ b/core/view/include/pulp/view/view.hpp
@@ -30,6 +30,12 @@ public:
         // pointer that the platform window host would dereference on
         // the next click.
         if (active_overlay_ == this) active_overlay_ = nullptr;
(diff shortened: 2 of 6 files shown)


danielraffel:
Codex P1 addressed in 6b25550c — `focused_input_` is now claimed/released on the standard mouseDown hit-test path (line 489) and on Tab focus navigation (line 682), in addition to the overlay click branch. This restores keyboard text entry for any TextEditor focused via plain click or Tab, not just overlay-routed clicks.

Issue #1714, opened 2026-05-10
fix(Label): set_multi_line(true) + setWhiteSpace('normal') doesn't actually wrap text


danielraffel:
## Summary

Calling `globalThis.setWhiteSpace(id, 'normal')` correctly invokes `Label::set_multi_line(true)` (per `widget_bridge.cpp:2457`), but the Label widget continues to paint text on a single line and truncate at the parent's width — it does not wrap.

## Repro

In `@pulp/react` JSX, with parent View having explicit width 100:

```jsx
<View style={{ width: 100 }}>
  <Label ref={el => globalThis.setWhiteSpace(el.id, 'normal')}>
    Color palette for bands and glow
  </Label>
</View>
```

Confirmed in stderr that `setWhiteSpace` runs successfully:

```
[wrap-ref] id="pr_2n" setWS=function
[wrap-ref] setWhiteSpace(pr_2n, 'normal') OK
```

But the Label paints as `Color palette for band` (truncated mid-word) with no wrap.

## Expected

The Label should wrap to fit the parent's allocated 100px width:

```
Color palette
for bands and
glow
```

## Likely cause

Either:
1. `set_multi_line(true)` sets the flag but `paint()` doesn't switch to multi-line layout.
2. `intrinsic_width()` returns the full single-line width regardless of `multi_line_`, so Yoga reserves enough single-line space and never asks for the wrapped height.
3. The measure callback doesn't honor Yoga's "fit to W" constraint mode.

## Spectr-side context

Surfaced while porting Spectr's Settings modal field hints to native bridge. Field labels live inside a 110px-wide column with `flexShrink: 0`, so children should wrap inside. With `whiteSpace: 'normal'` set explicitly via the dom-adapter's ref-callback, every hint Label still single-lines and clips ("Color palette for band…", "Halo intensity when ban…", etc.).

Repo: `/tmp/spectr-native-pump-fix/native-react/spectr-editor-extracted.js` line ~3577 (Field hint), `/native-react/dom-adapter.tsx` line ~1177 (synthetic Label ref-callback that calls setWhiteSpace).

Bridge function: `core/view/src/widget_bridge.cpp:2457` (setWhiteSpace).
Label widget: `core/view/include/pulp/view/widgets.hpp:78` (set_multi_line).

Related to but distinct from #1410 (white_space_nowrap surface) — that PR added the API; this issue is that the API doesn't actually wrap.

danielraffel:
## Hourly triage (automated)

[...]

PR #1743, opened 2026-05-10
fix(view): Label::set_multi_line(true) now actually wraps (#1714)

danielraffel:
## Summary

Fixes pulp #1714 — `Label::set_multi_line(true)` was a no-op for layout. Multi-line labels collapsed to a single line because:

1. `Label::intrinsic_height()` returned single-line height regardless of `multi_line_`
2. `yoga_measure` ignored width constraint when computing wrapped text height
3. Yoga measure-func was not registered when both intrinsic dims were 0
4. `set_multi_line()` did not invalidate layout

Now:
- `Label::wrap_lines(max_width)` does greedy word-wrap via global `TextShaper`
- `Label::measure_height_for_width()` returns `line_height × line_count`
- `yoga_measure` honors `widthMode` and dispatches to multi-line measure
- `set_multi_line()` invalidates layout from the root and triggers repaint
- `multiLine` prop wired through `@pulp/react` (types.ts + prop-applier)

Verified live in Spectr: Settings field hints now wrap to 2-3 lines fitting the 110px column.

## Test plan
- [x] 6 unit tests in `test/test_label_wrap.cpp` (12 assertions)
- [x] Existing test suite passes (`ctest --test-dir build`)
- [x] Live verification in Spectr standalone

Closes pulp #1714

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.87.3
+    VERSION 0.87.5
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/view/src/widgets.cpp
+++ b/core/view/src/widgets.cpp
@@ -255,6 +255,160 @@ float Label::intrinsic_height() const {
     return line_height_ > 0 ? line_height_ : effective_font_size * 1.4f;
 }
 
+std::vector<std::string> Label::wrap_lines(float max_width) const {
+    // pulp #1714 — greedy word-wrap into lines that fit within `max_width`.
+    // Uses the global TextShaper for accurate per-word advances; degrades
+    // to character-width estimation on non-Skia/non-HarfBuzz backends.
+    // Honors hard `\n` separators in the source text.
+    std::vector<std::string> lines;
+    if (text_.empty()) return lines;
+
+    float effective_font_size = font_size_;
+    if (!has_own_font_size_) {
+        if (auto inh = inheritable_font_size(); inh.has_value())
+            effective_font_size = inh.value();
+    }
+    float effective_letter_spacing = letter_spacing_;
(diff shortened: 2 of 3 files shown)


danielraffel:
Both Codex P2s addressed in db4c4b6f:

- **Font** — `wrap_lines` now measures with `font_family_` (falling back to "Inter" for the documented default), so Yoga height matches `paint()` for labels with custom typography.
- **Spacing** — wrap pass now captures the gap between tokens verbatim and restores it on non-breaking joins. `"foo  bar"` and leading indents survive. At wrap boundaries the inter-token gap drops, matching browser behavior where trailing whitespace before a line break is not painted.

3 new tests in test_label_wrap.cpp pin both fixes.

danielraffel:
Closing as obsolete — superseded by #1797.

**Why obsolete:**

- This PR added `Label::wrap_lines` + `Label::measure_height_for_width` to address Codex P2×2 on #1714 (font_family_ + authored-spacing preservation in width-wrap).
- #1797 (`feat(view): wire Label::paint to TextShaper::layout_with_lines`) merged after this was queued and took a different design: removed `wrap_lines` / `measure_height_for_width` entirely, replacing them with `TextShaper::layout_with_lines` gated on CSS `word-break: break-word | anywhere` + `bounds().width > 0`.
- The Codex P2 #1 finding (hardcoded "Inter") is now addressed in the new path — `Label::paint` resolves the family at line 378 (`font_family_.empty() ? "Inter" : font_family_`) and passes it through to `shaper.prepare(display_text, family, effective_font_size)` at line 475.
- The Codex P2 #2 finding (authored-spacing preservation) is now TextShaper's responsibility via `layout_with_lines`.
- Forward-porting this PR against current main produced compile errors because the out-of-line method definitions no longer match any declaration — `widgets.hpp` removed them.

**Residual gap:** `Label::intrinsic_width` at widgets.cpp:318 still hardcodes `"Inter"`. That's a separate, unrelated Codex P2 — will file a follow-up issue if it's still surfacing on real labels. Not blocking this closure.

No migration needed for consumers — the `Label::paint` width-wrap path is now opt-in via `word-break`.

PR #1754, opened 2026-05-10
fix(canvas2d): DOMMatrix scaleSelf/rotateSelf/inverse spec conformance (Codex P1×2 + P2 on #1730)

danielraffel:
## Summary

Three spec divergences in `_PulpCanvasMatrix` shim flagged by Codex review on PR #1730:

**P1 — `rotateSelf` takes degrees, not radians.** Per the [DOMMatrix spec](https://drafts.fxtf.org/geometry/#dom-dommatrix-rotateself), browser code calling `m.rotateSelf(90)` expects a quarter turn. Previous code treated the arg as radians. Now converts degrees → radians at the boundary. Omitted angle defaults to 0.

**P1 — `inverse()` returns NaN matrix for singular, not throw.** Per the [DOMMatrixReadOnly spec](https://drafts.fxtf.org/geometry/#dom-dommatrixreadonly-inverse), non-invertible matrices yield NaN components and `is2D = false`; they do **not** throw. Previous `TypeError` broke compat with callers relying on the standard non-throwing contract.

**P2 — `scaleSelf()` omitted args default to identity, not NaN.** Bare `scaleSelf()` now no-ops instead of producing NaN.

## Test plan
- [x] Rewrote 2 existing tests that documented the buggy behavior:
  - `rotateSelf 90deg` previously passed `Math.PI/2` — now passes `90` (degrees, per spec)
  - `inverse... singular throws` previously asserted throw — now asserts NaN matrix with `is2D=false`
- [x] 4 new spec-conformance tests:
  - `rotateSelf takes degrees` (90deg, 180deg)
  - `rotateSelf()` omitted defaults to 0 (no-op, not NaN)
  - `scaleSelf()` / `scaleSelf(N)` honor spec defaults
  - `inverse()` of singular returns NaN matrix with `is2D=false`
- [x] 96 cases in `pulp-test-canvas2d-shim` pass (273 assertions)

Co-ordinated via `planning/coverage-coordination.md`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.87.2
+    VERSION 0.87.3
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/view/js/web-compat-canvas.js
+++ b/core/view/js/web-compat-canvas.js
@@ -216,6 +216,10 @@ _PulpCanvasMatrix.prototype.multiplySelf = function(other) {
     return this;
 };
 _PulpCanvasMatrix.prototype.scaleSelf = function(sx, sy) {
+    // Codex P2 on #1730: spec says scaleX defaults to 1 when omitted,
+    // and scaleY defaults to scaleX when omitted. Without this, the
+    // bare scaleSelf() call multiplies by undefined and produces NaN.
+    if (sx === undefined) sx = 1;
     if (sy === undefined) sy = sx;
     this.a *= sx;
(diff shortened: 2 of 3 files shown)


danielraffel:
Both Codex P2s addressed in 2eaf1958:

- **All 16 components NaN** — `new _PulpCanvasMatrix(NaN×6)` only set the 2D aliases; `m13/m14/m23/m24/m31..m34/m43/m44` stayed at constructor-default identity. Now explicitly NaN-fills the remaining 10 entries so `toFloat32Array()` / `toFloat64Array()` return all-NaN.
- **toJSON honors is2D** — was hardcoded `is2D: true`. Now reads `this.is2D !== false` so a singular-inverse result correctly serializes `is2D: false`.

2 new tests pin both fixes (toFloat32Array all-NaN + toJSON is2D round-trip).

PR #1777, opened 2026-05-10
fix(html): restore :root pseudo-class for stylesheet matching (Codex P2 on #1773)

danielraffel:
## Summary

Closes Codex P2 finding on #1773. PR #1759 removed `:root` from `_matchesPseudoClass` entirely — but the same evaluator is called from `StyleSheet._applyTo`, not just from `document.querySelector`. So `:root { ... }` rules silently stopped applying. The most common casualty is CSS custom-property themes: `:root { --accent: red; }` no longer styled `document.body`.

The fix re-adds `:root` matching at the primitive level. An element matches when it's `__bodyElement__` (the stylesheet iteration root in this shim) or `__documentElement__` (real `<html>` semantics). The `querySelector` BFS path is unchanged — `:root` still can't be reached there because root isn't enqueued — but the primitive is now spec-conformant for both call sites.

## Why this matters

The Codex review on #1759 correctly flagged that the wholesale removal was too aggressive. The right level to disable `:root` was at the BFS traversal (which we never actually changed), not at the pseudo-class evaluator (which is shared between traversal and stylesheet application).

## Test plan

- [x] New regression test `:root pseudo-class matches body/documentElement for stylesheet themes [issue-1773]` in `test/test_widget_bridge.cpp` — creates a `StyleSheet({':root': {width: '321px'}})`, asserts body picks up the style and an inner div does NOT.
- [x] Re-ran existing `[issue-1737]` pseudo-class suite (8 cases, 59 assertions) — no regression.
- [ ] CI macOS local smoke + sanitizers (run on push)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ to its [GitHub Release](https://github.com/danielraffel/pulp/releases).
      next regen as long as they land in the right release's bullet block. See
      docs/guides/versioning.md § Release pipeline for the full end-to-end flow. -->
 
+<a id="v0811"></a>
+## [0.81.1] - 2026-05-10
+
+- fix(html): restore :root pseudo-class for stylesheet matching (Codex P2 on #1773) ([#1777](https://github.com/danielraffel/pulp/pull/1777))
+
 <a id="v100"></a>
 ## [1.0.0] - 2026-05-10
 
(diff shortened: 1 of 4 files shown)


danielraffel:
Superseded by #1779 (same Codex P2 fix landed at 21:57Z while this was queued on macOS smoke). Approach was equivalent — fix lives in `_matchesPseudoClass` so stylesheet path matches. Closing to drop the redundant queue slot.

Issue #1823, opened 2026-05-12
bug(import): non-runtime HTML parsers strip onclick/role/cursor before widget promotion sees them


danielraffel:
## Context

After the widget-promotion post-pass landed (closes follow-up to #1814 / task #84):

- `pulp::import_design::promote_interactive_frames` walks the parsed IR and re-types any `frame` that carries an interactive signal (`attributes["onclick"]` / `attributes["onClick"]`, `attributes["role"] == "button"`, or `style.cursor == "pointer"`) to `button`.
- The pass is source-agnostic, so it should benefit every importer once the parsers preserve the relevant inputs in the IR.

## Symptom

The smoke fixture:

```html
<div onclick="alert('hi')">click me</div>
<div role="button">aria button</div>
<div style="cursor:pointer">cursor button</div>
```

Imported via `pulp import-design --from stitch …` produces three `createLabel(…)` calls — **no promotions fire**. The IR walker prints no "Promoted N interactive frame(s)" line.

Same fixture imported via `parse_claude_html_with_runtime` (which walks the live DOM after React mount) **does** populate `IRNode::attributes` with the raw HTML attrs, so the post-pass promotes correctly. The unit test (`test_widget_promotion.cpp` / `pulp-test-widget-promotion`) covers that contract.

## Root cause

The non-runtime HTML/JSON parsers (`parse_stitch_html`, `parse_v0_tsx`, `parse_pencil_json`, and Figma JSON's attribute-collapse path) currently don't surface the source HTML attribute table into `IRNode::attributes`. They preserve enough to drive layout (id / class / style) but drop the interaction-relevant attrs.

`design_import.cpp`'s walker only inflates `attributes` from the parsed JSON object's `attrs` member; the stitch/v0/pencil bridges don't construct that member.

## Proposed fix

Extend each non-runtime parser to forward at least these attributes into `IRNode::attributes`:

- `onclick`, `onClick`, `onmousedown`, `onkeydown`
- `role` (for ARIA semantic claims)
- The `style.cursor` value already routes via `IRStyle::cursor`; that one's covered.

Once these survive parsing, the existing widget-promotion pass picks them up with zero additional code. No new IR shape required — only attribute fan-through.

Hot-files note: this touches `core/view/src/design_import.cpp` which is currently in Agent A's runtime-import work zone. File-and-park for now; pick up after #468 stabilizes.

## Verification

[...]

Issue #1814, opened 2026-05-11
pulp install: stale SDK cache silently shadows new releases — no version check, no version suffix on cached tarball


danielraffel:
## Bug

`pulp install` reuses a cached SDK tarball **without checking if it's current**, so the CLI silently ships an out-of-date SDK to consumers after a new release.

Reproducer (observed 2026-05-11 after v0.88.0 was published at 10:28 UTC):

```
$ pulp install
Installing Pulp SDK...
SDK (includes Skia) already cached at /Users/danielraffel/.pulp/cache/pulp-sdk-darwin-arm64.tar.gz

$ ls -la ~/.pulp/cache/pulp-sdk-darwin-arm64.tar.gz
-rw-r--r--@ 1 danielraffel staff 20244341 May  8 12:57 pulp-sdk-darwin-arm64.tar.gz
```

The cached tarball is from **May 8** (3 days before v0.88.0 was published). `pulp install` happily "installs" it without:
1. Checking against the latest published release tag
2. Validating an etag / sha256 against the release asset
3. Even just hitting `/releases/latest` and comparing dates

## Surface area

Directly affected: every consumer running `pulp install` between SDK releases. They get whatever was cached at first download, not the latest.

Indirectly affected: this is the upgrade pipeline for Spectr-class consumers. With v0.88.0 carrying the band-drawing / zoom / event-contract fixes from #1792 + the Panel/percent-radii fixes from #1802, users expecting `pulp install` + `pulp upgrade --install` to bring in those fixes will be stuck on the pre-cached version. They'll come away thinking "the fixes don't work for me" when really the bits never arrived.

## Additional context

- The cached tarball filename has **no version suffix** (`pulp-sdk-darwin-arm64.tar.gz`), so there's no way to tell from the filename alone whether it's stale. A version-pinned filename (`pulp-sdk-v0.88.0-darwin-arm64.tar.gz`) would let install fall back to a fresh download on any version mismatch.
- Adjacent: `pulp upgrade --notes --from 0.78.3 --to 0.88.0` returns: "(Migration index not yet ported to pulp-rs; no notes available.)" — separate gap, mention noted because consumers stepping over multiple versions get no migration guidance.

## Proposed fix

Minimal:
1. **Version-suffix the cached filename**: `pulp-sdk-vX.Y.Z-darwin-arm64.tar.gz`. Cache lookup by exact version → forces fresh download when target version doesn't match.
[...]

Issue #84, opened 2026-04-09
UI: Oscilloscope zero-crossing trigger for stable waveform display


danielraffel:
## Problem
The oscilloscope widget displays the raw buffer without triggering, so stable periodic signals appear to slide across the screen instead of holding still.

## Proposed
Add a zero-crossing trigger mode that aligns the displayed window to a rising-edge zero crossing. Lock the displayed waveform to the trigger point for periodic signals.

## Acceptance
- [ ] Oscilloscope widget has a trigger mode enum (free-run vs rising zero)
- [ ] Rising-zero mode produces a visually stable display for sine/square/etc.
- [ ] Default behavior unchanged (free-run) to avoid breaking examples
- [ ] Added to the scope test harness

PR #1824, opened 2026-05-12
feat(import-design): promote interactive frames to buttons (post-parse pass)

danielraffel:
External design tools (Figma, Stitch, v0, Pencil, Claude Design)
rarely emit <button> directly — the common pattern is
<div onClick={...}>, <div role="button">, or a styled div with
cursor:pointer. Without post-parse promotion the importer treats
those as static frames and the user's interactive intent is dropped:
Pulp ends up with a wall of frames where the original design has
clickable widgets.

This change adds a source-agnostic post-parse pass that walks the
IR once (after parse, before codegen) and re-types any
type == "frame" carrying an interactive signal to "button". Signals
are inspected in priority order:

  1. onclick / onClick attribute  — strongest signal
  2. ARIA role == "button"        — explicit semantic claim
  3. style.cursor == "pointer"    — weakest, opt-out via
                                    role="presentation"

Conservative by construction: only frames are promoted, never the
other direction; already-typed widgets (input / image / button)
are left alone so an <input onClick={...}> keeps its input type.
Idempotent — re-running the pass on already-promoted IR is a no-op.

Wired into the CLI pipeline in pulp_import_design.cpp between the
parse switch and codegen so every source benefits the moment its
parser preserves the relevant attrs.

Test: test/test_widget_promotion.cpp — 9 Catch2 cases / 36
assertions covering the four promotion paths (HTML lowercase
onclick, JSX camelCase onClick, ARIA role=button, cursor:pointer),
the role="presentation" opt-out, the already-typed guardrail, the
recursive walk, idempotence, and the signal-precedence pin. Builds
the tools/import-design source directly into the test binary so we
exercise the exact code path the CLI uses.

Known follow-up (#1823): the non-runtime HTML/JSON parsers
(parse_stitch_html, parse_v0_tsx, parse_pencil_json, parse_figma_json's
JSON-attrs path) currently strip onclick / role from the IR before
this pass sees them, so promotion only fires on the runtime-import
path (parse_claude_html_with_runtime) — the path Spectr re-import
actually uses. The post-pass is source-agnostic so it'll
automatically benefit every parser once #1823 lands.

Skill update: .agents/skills/import-design/SKILL.md gains a section
documenting the promotion priority order + the parser-strips-attrs
gotcha + how to read the new "Promoted N" stdout line.

Refs pulp #1814 and pulp-internal task #84. Filed pulp #1823 as the
parser-side follow-up.

Agent: B
Version-Bump: skip reason="importer-internal post-pass + new unit test + skill update — no SDK API or plugin surface change; the new widget_promotion translation unit is linked into pulp-import-design only"
Version-Bump: sdk=skip reason="no SDK API surface change — new helper is internal to pulp-import-design"
Version-Bump: plugin=skip reason="no plugin command/skill surface change"

### DIFF
--- a/.agents/skills/import-design/SKILL.md
+++ b/.agents/skills/import-design/SKILL.md
@@ -396,3 +396,23 @@ if (typeof globalThis.registerPointer === 'function') globalThis.registerPointer
 ```
 
 The cleaner long-term fix is for `@pulp/react`'s prop-applier to call `registerPointer` automatically when it sees any pointer-event prop (parallel to its existing `registerHover` wiring) — track that as a follow-up Pulp issue rather than an importer-side workaround if you encounter it on a fresh import.
+
+### 9. Post-parse widget promotion — `<div onClick>` → button
+
+`pulp::import_design::promote_interactive_frames` (in `tools/import-design/widget_promotion.{hpp,cpp}`) walks the IR once after parse + before codegen and re-types any `type == "frame"` carrying an interactive signal to `type == "button"`. Signal priority (highest → lowest):
+
+1. `attributes["onclick"]` / `attributes["onClick"]` — strongest.
+2. `attributes["role"] == "button"` — explicit ARIA semantic.
+3. `style.cursor == "pointer"` — weakest; opt-out via `role="presentation"`.
+
+Conservative on purpose: only frames are promoted; already-typed widgets (`input`, `image`, `button`) are left alone, so a designer who wrote `<input onClick={...}>` keeps the input.
+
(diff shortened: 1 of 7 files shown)


danielraffel:
Addressed both Codex P2s:

- **Stop descendant promotion**: walker now `continue`s after promoting, so clickable container + clickable child no longer produces nested buttons. New test pins it.
- **Iterative walk**: replaced the recursive lambda with a heap-backed `std::vector<IRNode*>` worklist. 5000-deep IR no longer stack-overflows; new tests cover deep + wide trees (12 cases / 46 assertions all green).

PR #1827, opened 2026-05-12
test(spectr): autonomous regression suite composing the #1792 event-bridge contracts

danielraffel:
## Summary

pulp #1792 landed six event-bridge fixes that silently broke Spectr (band-drawing, keys, global event fan-out, handler-throw resilience, registrar idempotency, wheel bubbling) and added six `[contract]` tests in `test_widget_bridge.cpp` that pin each invariant **in isolation**.

This PR adds `test/test_spectr_regression.cpp` — a complementary suite that composes the same six contracts into a Spectr-shaped mini scenario: a React-ish editor mount that registers handlers in the order Spectr's native-react loader does, then drives the bridge with a sequence of interactions that, pre-#1792, all silently failed.

The point: if a future PR keeps each unit-level contract passing but breaks their *interaction* (dispatch order, registrar-gate semantics, handler-wrap layering), this suite fails first.

## What the five tests pin

1. Left-click reaches the band-drawing handler as `e.button === 0` (W3C mapping).
2. Handler-throw doesn't poison subsequent dispatch — a throw on widget A doesn't drop a later event on widget B.
3. `window.addEventListener('keydown', fn)` sees every bridge keydown (global fan-out + W3C key strings together).
4. `registerPointer` / `registerWheel` are idempotent across re-mounts.
5. Full mixed sequence — throw + button-mapping + global key — composite pin.

Wired as `pulp-test-spectr-regression`, ctest-discovered.

## Test plan

- [x] `cmake --build build --target pulp-test-spectr-regression` builds clean.
- [x] `./build/test/pulp-test-spectr-regression` — 5 cases / 16 assertions all pass.
- [ ] Post-merge: `ctest -R spectr-regression` on CI passes on macOS + Linux + Windows.

Closes pulp-internal task #63.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1537,6 +1537,15 @@ add_executable(pulp-test-widget-bridge test_widget_bridge.cpp)
 target_link_libraries(pulp-test-widget-bridge PRIVATE pulp::view Catch2::Catch2WithMain)
 catch_discover_tests(pulp-test-widget-bridge)
 
+# Autonomous Spectr regression suite (pulp-internal task #63, follow-up to
+# pulp #1792). Composes the six [contract] invariants from
+# test_widget_bridge.cpp into a Spectr-shaped mini-scenario so a future
+# PR that keeps each unit-level invariant passing but breaks their
+# *interaction* still fails first.
+add_executable(pulp-test-spectr-regression test_spectr_regression.cpp)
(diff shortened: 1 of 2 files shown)


danielraffel:
Addressed Codex P2. Idempotence test now fires two wheel events on `editor-wrap` and asserts both `trace.wheels` and `trace.wraps_wheel` — if `registerWheel` ever stacks duplicate listeners, the deltaY list grows past 2 entries (or the counter past 2) and the test fails. 19 assertions across 5 cases all green.

PR #35, opened 2026-04-07
Add package manager README and removal guide

danielraffel:
Moves the README and REMOVAL.md from develop/package-manager to main, where all other package manager files already live.

PR #1828, opened 2026-05-12
feat(cmake): Debug-SDK perf-killer guard in PulpConfig.cmake

danielraffel:
## Summary

A Pulp SDK that was built with `CMAKE_BUILD_TYPE=Debug` ships debug-mode Skia, debug-mode audio framework, and debug-mode JS engine. Consuming it from a downstream plugin produces real-time audio so slow that buffer underruns happen on every reasonable host — a UX failure mode that's hard to attribute to "the SDK install is wrong" because the link succeeds and the plugin appears to work.

This PR wires a structural guard that refuses (or, with an opt-in, warns) when `find_package(Pulp)` resolves to a Debug-built SDK.

## What changed

- `CMakeLists.txt` — install step writes `${install-prefix}/sdk_build_type.txt` with the SDK's `CMAKE_BUILD_TYPE`.
- `tools/cmake/PulpConfig.cmake.in` — new guard reads the marker. `Debug` → `FATAL_ERROR` with remediation text; `PULP_ALLOW_DEBUG_SDK=ON` demotes to `CMake Warning` + proceed; missing marker (older SDK installs) or multi-config generators fall through silently.

## Test plan

- [x] `ctest -R "cmake-debug-sdk-guard"` — all 3 scenarios pass:
  - `debug-no-override` → expects `FATAL_ERROR` with the documented text.
  - `debug-with-override` → expects `CMake Warning` + exit 0.
  - `release` → expects silent + exit 0.
- Hermetic test extracts the live guard block from `PulpConfig.cmake.in` and runs it against a tempdir-fabricated SDK install, so future edits to the guard are auto-tracked.
- [ ] Post-merge: download a Release SDK release artifact, drop a `sdk_build_type.txt` with `Debug` into it, run `cmake -DCMAKE_PREFIX_PATH=...` from an example plugin and verify the `FATAL_ERROR` surfaces with the documented text.

This is the structural half of the Debug-SDK story. Pulp-internal #36 will document the plan to make a Debug SDK *actually usable* (for data-collection scenarios where real-time deadlines don't matter).

Closes pulp-internal task #35.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -808,6 +808,25 @@ endforeach()
 file(WRITE "${CMAKE_BINARY_DIR}/version.txt" "${PROJECT_VERSION}\n")
 install(FILES "${CMAKE_BINARY_DIR}/version.txt" DESTINATION ".")
 
+# SDK build-type marker (pulp-internal task #35).
+#
+# Records the CMAKE_BUILD_TYPE the SDK was built with so PulpConfig.cmake
+# can refuse — or at least loudly warn — when a downstream plugin tries
+# to consume a Debug-built Pulp SDK. Debug-built Skia + audio framework
+# blow real-time deadlines on every reasonable host, so consumers who
+# stumble into this on a local checkout deserve a clear "this SDK
+# install is unusable for plugin work" message instead of a silently
+# unusable plugin that drops audio.
+#
+# Multi-config generators (Visual Studio, Xcode) don't have a single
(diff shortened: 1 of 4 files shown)


danielraffel:
Addressed Codex P1 (case-insensitive build-type compare). Marker is now lowercased before the `STREQUAL` so `debug` / `DeBuG` / etc. all trigger the guard. Test grew two new scenarios: `debug-lowercase` and `debug-mixed-case` — both pass.

PR #1875, opened 2026-05-12
feat(import-validation): semantic-probe vector for runtime-import harness

danielraffel:
Pixel diff alone has documented blind spots in the Spectr import-validation
loop: a render that "looks right" can still have silent runtime failures
the histogram never sees. Codex review of the
spectr-validated-runtime-import-product-spec.md flagged this as a hard-gate
addition to the existing diff_against_reference{,_regions}.py path.

Adds tools/import-validation/semantic_probes.sh — a complement to the
pixel diff, scored over artifacts produced by an earlier Spectr launch:

  Probe 1 (runtime_import_err)
    __pulpRuntimeImportErr__ must be empty in the runtime log. The C++
    bridge's install_runtime_import_handlers() path (PR #1856) writes any
    Babel-transform / payload-eval / missing-host-primitive failure here.
    A render with partial React mount can still pass histogram diff
    because the chrome painted; this probe fails on any non-empty value.

  Probe 2 (runtime_import_trace)
    The runtime log must contain both phase=mounted AND phase=settled
    markers. Catches "render that bailed before useEffect / rAF queues
    drained" — pixels are fine, lifecycle never finished. Advisory by
    default (trace wiring is still landing on the Spectr side); pass
    --require-trace to enforce.

  Probe 3 (canvas_non_blank)
    Shells out to diff_against_reference_regions.py (PR #1871) --json
    and inspects central_canvas.blank_candidate. Catches "blank canvas
    behind intact chrome" — the chrome dominates the global histogram,
    overall similarity score passes, but the actual product surface is
    empty. Falls back to diff_against_reference.py whole-frame blank
    detection if the per-region script isn't yet present.

Constraints met:
  - Plain bash + python3 (no new deps beyond Pillow, already required).
  - Standalone: takes artifacts (log + screenshot), no Spectr launch.
  - Robust to missing log file (probes 1+2 skip, probe 3 still runs).
  - Robust to missing per-region diff script (probe 3 fallback).
  - JSON output (--json) for programmatic consumers.

Does NOT edit spectr-roundtrip.sh — scope kept tight to the new probe
script. The probes operate on artifacts the roundtrip already writes
(/tmp/spectr-rt-runtime.log + planning/screenshots/spectr-native-latest.png).

Verified scenarios (manual test fixtures, regions script + PR #1871
temporarily staged for the canvas-probe path):
  - All-good log + non-blank shot → 3 PASS, exit 0
  - Bad-err log → probe 1 FAIL with offending line surfaced
  - Missing-trace log + --require-trace → probe 2 FAIL listing missing phase
  - Blank shot → probe 3 FAIL with score + blank flag
  - Missing log entirely → probes 1+2 SKIP, probe 3 runs
  - Missing regions script → probe 3 falls back to whole-frame blank check
  - JSON output mode → clean machine-readable structure

Also adds tools/import-validation/README.md cataloguing the existing
scripts and documenting the probe / pixel-diff division of labor.

Version-Bump: skip reason="pure tooling — no shipped SDK/plugin/compat surface"

### DIFF
--- a/tools/import-validation/README.md
+++ b/tools/import-validation/README.md
@@ -1,0 +1,75 @@
+# tools/import-validation/
+
+Scripts for the Spectr import-validation harness — the closed loop that
+re-runs `pulp import-design` against `spectr/resources/editor.html`,
+relaunches the native Spectr build, captures the live render, and compares
+it against the canonical browser screenshot.
+
+See `planning/spectr-validated-runtime-import-product-spec.md` (private
+submodule) for the full design.
+
+## What's in here
+
+| Script | Purpose |
+|--------|---------|
+| `spectr-roundtrip.sh` | The full A→D loop: re-import editor.html → rebuild Spectr → launch → capture → diff. Top-level entry point for "did my Pulp fix narrow the gap?" |
+| `diff_against_reference.py` | Single-score histogram + pixel-distance diff between two PNGs. Used by `spectr-roundtrip.sh` step 5. |
+| `diff_against_reference_regions.py` | (PR #1871) Per-region masked diff — fails on the FIRST broken sub-region instead of averaging the whole frame. Recommended over the single-score variant once landed. |
+| `semantic_probes.sh` | **Semantic-probe vector** — pixel-diff complement. Asserts no soft runtime-import error, lifecycle reached `mounted`+`settled`, and the canvas region actually painted. See below. |
+| `check_label_coverage.sh` | Structural label-coverage check — string-match expected reference labels against the imported IR. |
+| `reference-labels-spectr.txt` | Ground-truth list of UI labels that must appear in any successful Spectr import. |
+
+## Semantic probes
+
+`semantic_probes.sh` is the codex-required addition to the pixel-diff
+pipeline (planning spec, 2026-05-12). Pixel diff alone has documented
+blind spots — a render that "looks right" can still have silent runtime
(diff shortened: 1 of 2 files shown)


danielraffel:
Heads-up: Shipyard's macOS lane reports `Stage 'test' failed` on this PR across three sequential attempts (`sy-20260512-fd3294`, `sy-20260512-5286d5`, `sy-20260512-f4c6ed`). The single failing test is **#3807 `PluginScanner scan runs without crash`** — a SEGFAULT inside the system-wide installed-plugin scan path (third-party VST3s on the host: HyperPrism, SWAM, Bendr, etc.).

The crash is unrelated to this PR:
- This PR only adds two files under `tools/import-validation/` (a bash probe script + a README); zero C++ changes, zero impact on `core/host/` or `PluginScanner`.
- 4814 of 4815 tests pass on the same run.
- Test #3807 calls `scanner.scan(opts)` which dlopen's every installed system plugin; the SEGFAULT comes from one of those vendor binaries, not from Pulp code under test.
- Adjacent shipyard runs from other branches today (`sy-20260512-cdd6b3`) passed 0/4808, so the failure is sensitive to current installed-plugin state, not branch content.

Leaving the PR open for the maintainer to decide between merging through the environmental flake or investigating the scanner crash separately. All non-required GHA checks are green; the per-PR version + skill-sync gate passed.

PR #2068, opened 2026-05-15
feat(mcp): ship pulp-mcp in tarball + per-tool feature detection (#2067, #2070)

### LINKED ISSUE #2067: Release: ship pulp-mcp (and pulp-scan-worker?) in CLI release tarballs
## Problem

`pulp-mcp` is NOT shipped in the release tarball, so users who install via the official one-liner can't use the Pulp MCP server out of the box.

```bash
curl -fsSL https://www.generouscorp.com/pulp/install.sh | sh
```

The `pulp-darwin-arm64.tar.gz` (and the other 4 platform tarballs) contain:

```
pulp
pulp-cpp
libwgpu_native.dylib
```

But NOT `pulp-mcp`. Then when you open the Pulp repo in Claude Code, `/mcp` shows:

```
Pulp MCP Server
Status:  ✘ failed
Command: ${CLAUDE_PLUGIN_ROOT}/tools/mcp/pulp-mcp-launcher
```

The launcher's resolution order is:
1. `$REPO_ROOT/build/tools/mcp/pulp-mcp` (source-tree build)
2. `pulp-mcp` on `$PATH` (installed binary)

A user with only the installed CLI has neither — they'd have to clone the source and run `cmake --build build --target pulp-mcp`, which is a ~10-min cold build because `pulp::mcp` transitively pulls in Skia / pulp::view / etc.

## Repro

```bash
curl -fsSL https://www.generouscorp.com/pulp/install.sh | sh
source ~/.zshrc
ls ~/.pulp/bin/    # → pulp, pulp-cpp, libwgpu_native.dylib (no pulp-mcp)
which pulp-mcp     # → not found
# Open Pulp repo in Claude Code → /mcp → "Status: ✘ failed"
```

## Fix

Bundle `pulp-mcp` (and likely `pulp-scan-worker` — same boat, only ships in some installs) alongside `pulp` and `pulp-cpp` in `pulp_${PLATFORM}.tar.gz`.

Two-line change to `tools/scripts/package_cli.py` to accept a `--mcp-binary` arg and copy it into the tarball, plus the corresponding wiring in `.github/workflows/release-cli.yml` Package CLI steps:

```yaml
# .github/workflows/release-cli.yml — Package CLI (Unix), around line 162
- name: Package CLI (Unix)
  if: runner.os != 'Windows'
  run: |
    python3 tools/scripts/package_cli.py \
      --binary build/pulp \
      --cpp-binary build/tools/cli/pulp-cpp \
      --mcp-binary build/tools/mcp/pulp-mcp \      # NEW
      --build-dir build \
      --platform ${{ matrix.platform }} \
      --out pulp-${{ matrix.platform }}.tar.gz
  shell: bash
```

Same for the Windows step (`build/tools/mcp/Release/pulp-mcp.exe`).

## Smoke

The existing PR #395 smoke matrix already loops `for ART in pulp pulp-cpp; do "$INSTALL_DIR/$ART" help; done` (release-cli.yml line 377). Extend it to `for ART in pulp pulp-cpp pulp-mcp; do …; done` — verifies the binary actually launched on a clean runner before publishing.

## Related

- The MCP launcher (`tools/mcp/pulp-mcp-launcher`) is correct as written — it falls back to PATH. Once `pulp-mcp` is on PATH after install, the launcher resolves it.
- `min_cli_version: "0.31.0"` on the marketplace plugin is satisfied by every release; this is purely about which binaries make it into the tarball.
- `pulp-scan-worker` may have the same gap — worth auditing.

## Acceptance

- [ ] `tar tzf pulp-darwin-arm64.tar.gz` shows `pulp-mcp` alongside `pulp` and `pulp-cpp`
- [ ] Fresh `curl … install.sh | sh` install on a clean machine results in `~/.pulp/bin/pulp-mcp` being executable
- [ ] `pulp-mcp --help` (or whatever the MCP server's bootstrap arg is) runs cleanly after install
- [ ] Pulp MCP Server status flips to ✓ in Claude Code's `/mcp` after install + reconnect
- [ ] Optionally: do the same for `pulp-scan-worker` if confirmed missing

danielraffel:
## Summary

Closes #2067 (tarball missing `pulp-mcp`) and #2070 (per-tool feature detection in pulp-mcp).

### #2067 — ship `pulp-mcp` in the CLI tarball
- `package_cli.py` gains `--mcp-binary`; `release-cli.yml` strips and bundles `pulp-mcp` on all 5 platforms; smoke matrix runs `pulp-mcp --version` (the only short-circuit flag that exits cleanly without consuming the JSON-RPC stdin loop).
- `serverInfo.version` was hard-coded `"0.1.0"` and never moved. Now tracks `PROJECT_VERSION` via `tools/mcp/pulp_mcp_version.h.in`.
- `pulp doctor` resolves `pulp-mcp` from `$PATH` → `~/.pulp/bin/` → source-build (Windows multi-config probe included), requires `--version` to actually emit the expected contract string before declaring success.
- Docs document "CLI first, plugin second" install order; `install.sh` / `install.ps1` log the binary landing.
- `cmd_create.cpp` now honors the `optional` flag on doctor pre-flight rows so the new optional `pulp-mcp` row doesn't break `pulp create` on fresh machines.

### #2070 — per-tool `min_sdk_version` + `pulp_compat` introspection
The plugin, `pulp-mcp`, and a project's pinned SDK are three independently-versioned surfaces. A user can have a newer `pulp-mcp` on PATH while editing a project pinned to an older SDK.

- New `TOOL_MIN_SDK_TABLE` in `pulp_mcp.cpp` lets each tool declare a minimum project SDK floor. Currently empty — all existing tools default to `"0.0.0"` (no floor), so behavior is unchanged. New tools opt in by adding a row.
- On every `tools/call`, the dispatcher reads the project's pinned SDK from `pulp.toml` `sdk_version` (preferred) or `CMakeLists.txt` `project(... VERSION ...)`, compares via semver triple, and returns a structured `isError: true` payload (`{error: "sdk_too_old", tool, required_sdk, project_sdk}` + LLM-readable text mentioning `pulp project bump`) when the project is too old. Fails OPEN on parse failure so a malformed `pulp.toml` can't strand the user.
- New `pulp_compat` tool reports `pulp_mcp_version`, `mcp_protocol_version`, `project_sdk`, and the per-tool `min_sdk` map so plugins can pre-filter their visible tool list at startup.
- Launcher stays version-tolerant; `pulp doctor` stays advisory-only. The fix lives per-tool, not at the connection boundary.

## Why not also ship `pulp-scan-worker` / `pulp-import-design` / etc?

- `pulp-scan-worker` already ships in the SDK tarball (`install(TARGETS)`).
- `pulp-import-design`, `pulp-design-debug`, `pulp-screenshot` are project-local build artifacts — only useful inside a project's build dir, not from `~/.pulp/bin/`.

## Codex review fixes (in-PR)

- **P2 (line 2039)** — Windows multi-config probe paths in `pulp doctor` advisory row.
- **P2 (line 2046)** — require successful `--version` execution (not just `fs::exists`) before declaring the row passed; surface a distinct ⚠ when found-but-unrunnable.

## Test plan

- [x] `python3 tools/scripts/test_package_cli.py` — 4 new tests for `--mcp-binary`. All 20 pass.
- [x] `./build/test/pulp-test-mcp-server` — 7 new test cases for compat (`compare_semver`, `min_sdk_for_tool`, `pulp_compat` with/without project, `compat_error_payload`, `parse_cmake_project_version`, `parse_pulp_toml_sdk_version`). All 140 assertions pass.
- [x] `pulp-mcp --version` outputs `pulp-mcp 0.102.0`.
- [x] `pulp_compat` returns the four expected fields and the per-tool matrix.
- [x] `pulp doctor` advisory row works in 3 states: found-and-runs (`✓`), found-but-broken (`⚠` naming path), missing (`⚠`).
- [x] `cmd_create.cpp` no longer blocks on missing pulp-mcp.
- [x] `python3 tools/scripts/check_cli_mcp_parity.py --mode=report` clean (`pulp_compat` added to `mcp_only`).
- [x] `python3 tools/scripts/docs_noise_lint.py --mode=report` clean.
- [x] `python3 tools/scripts/version_bump_check.py --base origin/main --mode=report --require-bump-for-fix-feat` clean.
- [x] Shipyard local macOS validation: all green.
- [ ] GitHub-hosted Linux/Windows/sanitizer matrix.
- [ ] Post-release verification (manual, after this lands and a release cuts):
  - `rm -rf ~/.pulp/ && curl install.sh | sh` → `~/.pulp/bin/pulp-mcp` exists + executable
  - `pulp upgrade --install` carries `pulp-mcp` forward
  - `claude plugin install pulp` results in `/mcp` showing "Pulp MCP Server: ✓ ok"


danielraffel:
Both Codex P2 comments addressed in 4ed2b0555:

**P2 (line 2039) — Windows multi-config probe:** `run_doctor_checks()` now probes `build/tools/mcp/{Release,Debug}/pulp-mcp.exe` in addition to the single-config path under Windows. Matches the layout release-cli.yml already packages from.

**P2 (line 2046) — Require successful execution:** The check walks candidates in priority order and only declares success after `--version` actually emits the contract string `pulp-mcp <semver>`. A path that exists but produces unexpected output (stale binary, missing dylib, no execute bit, codesign tear) now renders as a distinct `⚠` row that names the failing path, so the user can investigate runtime issues instead of chasing a generic "missing" diagnostic.

Manually verified on macOS:
- found + clean: `✓ pulp-mcp — <path> (pulp-mcp 0.101.7)  [CLI 0.101.7]`
- stale binary on PATH: `⚠ pulp-mcp — /tmp/pulp-mcp found but --version did not return the expected pulp-mcp <semver> line`
- missing: `⚠ pulp-mcp — not found — the Claude Code plugin's MCP server will fail`

danielraffel:
Per-tool feature detection from #2070 also landed in this PR (commit `861e7a01c`), so the three-channel skew story is now end-to-end:

**Build version axis**
- `pulp-mcp --version` and MCP `initialize`'s `serverInfo.version` track `PROJECT_VERSION` via `tools/mcp/pulp_mcp_version.h.in` (was hard-coded `"0.1.0"` and never moved pre-PR).
- `pulp doctor` advisory row reports the binary's path + `--version` output alongside the CLI version so users see drift.

**MCP wire protocol axis**
- `protocolVersion: "2024-11-05"` from the upstream MCP spec, independent of Pulp's build version. Exposed via the new `pulp_compat` introspection tool.

**Project SDK skew axis (the new bit)**
- `TOOL_MIN_SDK_TABLE` in `tools/mcp/pulp_mcp.cpp` lets each tool declare a `min_sdk_version` floor. Currently empty — every existing tool defaults to `"0.0.0"` (no floor) so behavior is unchanged. New tools opt in by adding a row when they begin to depend on a specific SDK API.
- On every `tools/call`, the dispatcher resolves the project's pinned SDK from `pulp.toml` `sdk_version` first, then `CMakeLists.txt` `project(... VERSION ...)`, compares via semver triple, and returns a structured `isError: true` payload when the project is too old:
  ```json
  {
    "isError": true,
    "structuredContent": {"error":"sdk_too_old","tool":"pulp_future_tool","required_sdk":"0.110.0","project_sdk":"0.99.0"},
    "content": [{"type":"text","text":"Tool `pulp_future_tool` requires project SDK >= 0.110.0; this project is pinned to 0.99.0. Bump via `pulp project bump`..."}]
  }
  ```
- Fails OPEN on parse failure (malformed `pulp.toml`, missing project root, unparseable semver) — a launcher-side hard gate would strand users; the policy is per-tool tolerance only.
- New `pulp_compat` introspection tool reports `pulp_mcp_version`, `mcp_protocol_version`, `project_sdk`, and the full `tool_min_sdk` map so plugins can pre-filter their visible tool list at startup.

7 new Catch2 cases cover `compare_semver` (numeric vs lex compare trap), `min_sdk_for_tool` default, `pulp_compat` with/without project root, `compat_error_payload` shape, `parse_cmake_project_version`, and `parse_pulp_toml_sdk_version` (including the `min_sdk_version` substring trap).

Closes #2070 in the same PR.

PR #2091, opened 2026-05-15
feat(cli): SDK-update UX — auto-update default, easy pin opt-out (#2087)

### LINKED ISSUE #2087: SDK-update UX: surface newer SDK availability + make install.sh suggest `pulp sdk install`
## Problem

A user runs `curl -fsSL https://www.generouscorp.com/pulp/install.sh | sh`, then `pulp version` shows v0.101.5 — looks current. But any project (e.g. Spectr) still links against whatever SDK was previously installed (e.g. v0.91.0-local from a months-ago `setup.sh`), silently missing every framework fix in between.

**Reproduction (real user session 2026-05-15):**

1. Fresh `curl … install.sh` puts pulp v0.101.5 in `~/.pulp/bin/`
2. `~/.pulp/sdk/` already contained one ancient install (0.91.0-local from May 12)
3. Spectr CMake `find_package(Pulp 0.87.0 REQUIRED)` → resolves to the 0.91 SDK
4. `pulp doctor --versions` from inside Spectr says:
   ```
   CLI:      v0.101.5
   SDK:      v0.91.0   (project)
   Skew checks: OK — CLI v0.101.5 is compatible with project SDK v0.91.0
   All checks passed.
   ```
5. User has no idea Spectr is missing 6 SDK versions of fixes (#1984 viewport-fit, #2010 ESC, #2011 SVG curves, fontconfig, mcp tarball, etc.)

## Design (current stage: solo user / frequent updates)

**Default behavior shifts to auto-update.** Pinning becomes opt-in, not opt-out. Rationale: at our current stage the cost of *missing* updates is high (this is a recurring real-user problem) and the cost of a *surprise* update is low (we ship the changes, we know what's in them). Once the user base grows and plugins start shipping to end-users, flip defaults to pinning-by-default — that's a two-line change in `pulp create` and the import-design banner copy.

### 1. `install.sh` installs CLI + latest SDK in one shot

Today: CLI only. Closing message says nothing about the SDK.
Proposed: `install.sh` runs `pulp sdk install` automatically after the CLI lands. Print:
```
Installed pulp v0.101.5 CLI + SDK at ~/.pulp/sdk/0.101.5/
```

### 2. `pulp upgrade` upgrades both CLI and SDK by default

Today: CLI only.
Proposed: default behavior fetches latest CLI **and** latest SDK. Add `--cli-only` for the rare opt-out (e.g., user explicitly wants to keep the project's current SDK paired with a newer CLI).

### 3. `pulp build` / `pulp import-design` etc. surface a single banner

When the project's resolved SDK is behind the latest installed (or the latest available remote), print one line:
```
$ pulp import-design --from claude --file editor.html
ℹ️  Spectr is using SDK v0.91.0 (resolved from ~/.pulp/sdk/0.91.0-local).
   SDK v0.102.0 is available with fixes relevant to import:
     #1984 viewport-fit, #2011 SVG curves, #2010 ESC, …
   Upgrade: pulp upgrade        # gets latest CLI + SDK + rebuilds
   Stay:    pulp project pin 0.91.0   # silence this notice
   …continuing with 0.91.0…
```

Suppressible via `~/.pulp/config.toml` (`[ui] sdk_update_notice = "off"`) or a per-project flag in `pulp.toml`.

### 4. `pulp create` defaults to floating mode

Writes `sdk = "latest"` to the new project's `pulp.toml`. Project tracks latest installed unless the user later runs `pulp project pin <ver>`. New projects start "current" and stay current.

### 5. `pulp project pin <ver>` is the explicit opt-out

Writes an exact version to `pulp.toml`, turns off auto-update for the project. The current `pulp project bump` flow already does most of this; the rename to `pin` (or add as alias) signals intent — "stop drifting on this project."

### 6. Show what changed when auto-updating

When `pulp upgrade` (or a command-driven auto-update) crosses minor versions, link to the changelog or print the high-level summary:
```
✓ SDK 0.91.0 → 0.102.0 installed.
  Changes you might care about:
    • viewport-fit: editor windows now letterbox to design viewport (#1984)
    • SVG <path> renders cubic curves (#2011)
    • ESC closes ComboBox even with stolen focus (#2010)
  Full changelog: pulp docs changelog
```

## Inspiration

| Tool | Lesson |
|------|--------|
| Rustup | Per-project `rust-toolchain.toml` pin works; **default** is "latest stable, sticky" |
| Deno | One-line "newer available" nag on every invocation |
| Cargo | `cargo update` is explicit; `cargo build` never auto-updates |
| Homebrew | `brew upgrade` updates everything by default |
| **Pulp (this proposal)** | Hybrid: install + upgrade auto-update, build surfaces nag, pin is opt-in |

## Migration path to pinning-default (when user base grows)

Two changes flip the default:
- `pulp create` writes `sdk = "0.X.Y"` (exact) instead of `sdk = "latest"`
- The banner in #3 changes from "we auto-updated" to "v0.X.Y is available; `pulp project bump` to upgrade"

Existing floating projects keep floating unless re-pinned. Nothing breaks.

## Acceptance

- `install.sh` installs CLI + latest SDK in one invocation; closing message mentions both versions
- `pulp upgrade` updates both by default; `--cli-only` works
- `pulp build` / `pulp run` / `pulp import-design` / `pulp ship` / `pulp create` print a single-line "newer SDK available" banner when behind, suppressible
- `pulp create` writes `sdk = "latest"` (floating) by default; `--pin` flag pins at create-time
- `pulp project pin <ver>` exists (or aliases `pulp project bump`) as the explicit pin
- `pulp doctor --versions` lists latest-installed AND latest-remote alongside project pin
- `pulp sdk available` (or `pulp sdk list --remote`) exists
- `pulp sdk install --help` documents `install` subcommand

## Non-goals

- Forcing pins on existing projects (only `pulp create` writes a default)
- Auto-updating a project's pinned SDK (pinning is sacred when the user has explicitly pinned)
- Adding `compatibility_with` semver metadata to releases (future work; needed for "safe auto-track within compat range" — but not required for solo-user stage)

Filed from a real user session 2026-05-15.

danielraffel:
## Summary

Closes #2087. Plus a piggyback CMake fix that unblocks downstream consumers (including Spectr) from rebuilding against v0.101.5+ on Linux/Android.

### Piggyback: `lib/cmake/Pulp/PulpUtils.cmake` doesn't reach into the Pulp source tree

PulpUtils.cmake's downstream-facing helpers were resolving paths via `${CMAKE_SOURCE_DIR}`, which in `find_package(Pulp)` context points at the *consumer's* source tree — not Pulp's. Every Linux/Android consumer's configure step failed:

```
CMake Error at .../lib/cmake/Pulp/PulpUtils.cmake:997 (include):
  include could not find requested file:
    <consumer-dir>/tools/cmake/PulpLinkFontconfig.cmake
```

Fix:
- Inside `_pulp_add_standalone` (a function body): use `CMAKE_CURRENT_FUNCTION_LIST_DIR` (CMake 3.17+; Pulp pins 3.24). `CMAKE_CURRENT_LIST_DIR` inside a function body resolves to the **caller's** dir, not the function's defining file — a classic CMake gotcha that bit on first attempt.
- Top-level fallbacks (`_PULP_FORMAT_SOURCE_DIR`, AUv3 templates, iOS AUv3 templates): probe in-tree layout first (`../../core/...`), then installed-tree layout (`../../../src/...`), then empty so plugin targets fail with a clear "missing source" error instead of silently picking up the consumer's tree.
- Root CMakeLists.txt: `PulpLinkFontconfig.cmake` now ships in the `install(FILES)` block alongside `PulpUtils.cmake` at `${CMAKE_INSTALL_LIBDIR}/cmake/Pulp/`.

New CI gate `.github/workflows/install-consumer-smoke.yml`:
- Installs Pulp to a temp prefix on macos-15 + ubuntu-24.04
- Configures a minimal downstream `find_package(Pulp)` + `pulp_add_plugin(...)` project
- Greps the installed CMake config files for `${CMAKE_SOURCE_DIR}/tools/...` or `.../core/...` (the exact bug pattern) and fails loudly if any future commit reintroduces the class

Verified end-to-end locally on macOS: install → consumer configure → "Pulp::view target resolved." ✓

### #2087: Auto-update default, easy pin opt-out

The Spectr-class incident: `curl install.sh | sh` puts a fresh CLI on top of whatever stale SDK the cache had — silently missing every framework fix in between. Fix: default to floating-SDK mode, make pinning opt-in.

- **`pulp create`** writes `sdk_version = "latest"` by default. `--pin` opts into exact-version pinning at create time.
- **`pulp project pin`** is the new primary command name; `bump` survives as a deprecated alias for one minor release. **`pulp project unpin`** (new) flips a project back to floating.
- **`pulp upgrade`** fetches CLI + matching SDK by default; `--cli-only` opts out. In a pinned project, the SDK swap leaves the pin alone with a clear notice ("Project X stays on pinned SDK Y.Y.Y…").
- **`pulp doctor --versions`** appends a "Local SDKs" section: newest installed under `~/.pulp/sdk/`, project pin status (floating vs pinned), cached latest remote, "newer SDK available" hint when behind.
- **`install.sh`** runs `pulp sdk install` after the CLI install so a fresh `curl install.sh | sh` leaves the user with a coherent CLI + SDK pair. Closing message names both versions and surfaces the `pulp project pin` command.
- **`read_sdk_version`** resolves the explicit `"latest"` marker to the newest installed SDK; new helpers `is_floating_sdk` / `read_raw_sdk_version` / `newest_installed_sdk` so callers don't open-code the comparison.

## Non-goals (per #2087 + deferred to follow-ups)

- **Auto-install-on-first-need** across every command (`pulp build`/`run`/`import-design`/`ship` in an unpinned old project). Needs config-knob plumbing across every SDK-resolution site; bigger than this PR's scope.
- **"Newer SDK available" one-line banner** on every command. Needs best-effort network call infrastructure with 1s timeout + a global suppress knob.
- **`pulp sdk available`** subcommand. Needs a new `fetch_all_releases` API alongside the existing `fetch_latest_release`.
- Existing projects are NOT auto-rewritten to floating. Only newly-created projects get the new default; existing `pulp.toml` files are untouched (pinning is sacred).

`pulp sdk install --help` is already documented in `cmd_sdk.cpp` — the recollection that it was missing turned out to be a no-op.

## Test plan

- [x] `ctest -R "cmd_project reports help|cli common project metadata"` — 2/2 pass
- [x] `ctest -R "[issue-2087]"` — pin/unpin help + unpin round-trip (43 assertions)
- [x] `pulp doctor --versions` shows local + remote SDKs + pin status
- [x] `pulp project --help`, `pulp project pin --help`, `pulp project unpin --help`, `pulp project bump --help` all render
- [x] Downstream consumer smoke (local): install → `find_package(Pulp)` → `Pulp::view` resolved
- [x] Grep guard for `CMAKE_SOURCE_DIR/tools/...` in installed configs — clean
- [ ] Shipyard macOS local validation
- [ ] `install-consumer-smoke.yml` runs on macos-15 + ubuntu-24.04

### DIFF
--- a/.agents/skills/ci/SKILL.md
+++ b/.agents/skills/ci/SKILL.md
@@ -1063,6 +1063,25 @@ a PR, and let the drift-check workflow confirm the plan. Then mirror the
 change in the GitHub ruleset UI (or reapply via `gh api PUT`). Never edit
 the live ruleset in isolation — the next scheduled drift run will fail.
 
+### Install consumer smoke (`install-consumer-smoke.yml`)
+
+Pulp #2087 piggyback. Catches the class of bug where in-tree builds
+work but installed-SDK consumers break at configure time. Runs on
+macos-15 + ubuntu-24.04: builds Pulp, `cmake --install`s it to a temp
+prefix, then configures a minimal downstream `find_package(Pulp)` +
+`pulp_add_plugin(...)` project against that prefix. Failures here
+match what a real downstream (e.g. Spectr) would hit.
+
+Defense-in-depth guard: greps the installed CMake config files for
+`${CMAKE_SOURCE_DIR}/tools/cmake/...` or `${CMAKE_SOURCE_DIR}/core/...` —
+those patterns inside files that ship in the SDK tarball resolve to
+the *consumer's* source tree at find_package time, never Pulp's.
+Inside a function body in `tools/cmake/PulpUtils.cmake`, use
+`CMAKE_CURRENT_FUNCTION_LIST_DIR`; at top level of a config file,
+use `CMAKE_CURRENT_LIST_DIR`. The two existing helpers paths
+(`_pulp_add_standalone` for fontconfig, top-level fallbacks for
+`_PULP_FORMAT_SOURCE_DIR` etc.) demonstrate the pattern.
+
 ## Versioning & Skill-Sync gates (Layer 3)
 
 `pulp pr` orchestrates the full shipping flow. CI enforces three gates on every PR to `main`:
--- a/.agents/skills/cli-maintenance/SKILL.md
+++ b/.agents/skills/cli-maintenance/SKILL.md
@@ -193,7 +193,7 @@ pulp #709 / `--from claude` is the worked example.
 - [ ] Search CLAUDE.md
 - [ ] Run sync check
 
-## `pulp project bump` shell helpers
+## `pulp project pin / unpin / bump` shell helpers
 
 `tools/cli/cmd_project.cpp` shells out to `git` / `cmake` from unit-tested
 helper paths, including Windows CI. Keep output redirection platform-aware:
@@ -202,6 +202,35 @@ POSIX uses `/dev/null`, but Windows `cmd.exe` needs `NUL`. Do not add raw
 the local null-redirection helpers instead. Otherwise Windows Namespace can
 misreport clean/dirty git state or fail origin-main probes even though the
 same tests pass on macOS/Linux.
+
+### pin / unpin / floating SDK mode
+
+`pin` is the primary command name; `bump` survives as a deprecated alias
+through `cmd_project`'s dispatch (`if (sub == "pin" || sub == "bump")`).
+When adding a new project subcommand, add the canonical name AND keep
+any old alias for one minor release — existing scripts and skill examples
+break otherwise.
+
+`pulp project unpin` rewrites `pulp.toml`'s `sdk_version` to `"latest"`
+in-place (single-line value swap, preserving surrounding TOML
+structure and comments). Do NOT delete the field — downstream tooling
+that greps for it loses a clear signal. The dispatch entry lives at
+the same spot as `pin`/`bump`/`undo` and shares `find_bumpable_project_root_from`.
+
+`sdk_version = "latest"` is the floating-SDK marker. Resolution happens
+in `read_sdk_version()` (cli_common.cpp) — `"latest"` becomes the
+newest installed version under `~/.pulp/sdk/<x.y.z>/`, falling back to
+`PULP_SDK_VERSION` when none are installed. Callers that need to
+distinguish the floating marker from a real semver use
+`read_raw_sdk_version()` + `is_floating_sdk()`. Don't open-code the
+"latest" comparison anywhere — both helpers are exported from
+cli_common.hpp so the comparison stays in one place.
+
+`pulp create` writes `sdk_version = "latest"` by default
+(cmd_create.cpp). The `--pin` flag opts into exact-version pinning
+at create time. Both code paths print a discoverable post-create
+message about `pulp project pin` so users learn the opt-in command
+without hunting.
 
 ## `pulp pr` — shim over `shipyard pr`
 
--- a/.agents/skills/upgrade/SKILL.md
+++ b/.agents/skills/upgrade/SKILL.md
@@ -69,6 +69,39 @@ move its SDK pin. It operates on the active project or the registry
 (`--all`); it does not upgrade the global CLI and it refuses to treat
 the Pulp source checkout as a consumer project.
 
+### `pulp upgrade` updates BOTH CLI and SDK by default
+
+Pre-#2087 `pulp upgrade` only swapped the CLI binary. Users who ran it
+then built a project that still resolved its (months-old) SDK and
+silently missed every framework fix in between (the Spectr incident
+2026-05-15). Post-#2087:
+
+- Default: fetch latest CLI binary + run `pulp sdk install --version
+  <new>` immediately after the swap, so the matching SDK lands at
+  `~/.pulp/sdk/<new>/` in the same invocation.
+- `--cli-only` flag keeps the pre-#2087 behavior for the rare case
+  where a user wants the new CLI paired with their current SDK.
+- The SDK install is best-effort: a transient network failure logs a
+  warning and tells the user to retry with `pulp sdk install`. The CLI
+  swap is not rolled back — the user can always retry SDK install.
+- When run from inside a project whose `pulp.toml` pins an explicit
+  SDK version, the SDK install happens (so the new version is
+  available globally) but the project's pin is left ALONE. A clear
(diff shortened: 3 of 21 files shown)


danielraffel:
Both Codex P2 comments addressed in afcd80d65, plus one additional fix surfaced by `/mcp` reproduction:

**P2 (install.sh line 136) — preserve SDK install exit code:**
The `if pulp sdk install | sed; then` pattern was masking failures — `set -e` doesn't help inside a pipeline and `set -o pipefail` wasn't set, so the pipeline status came from `sed` (always 0) regardless of whether `pulp sdk install` succeeded. Replaced with `${PIPESTATUS[0]}` capture so the success message is gated on the real return code. CLI install still doesn't fail-fast on SDK failure (user has an actionable retry path), but the warning now prints the actual exit code.

**P2 (install-consumer-smoke.yml line 111) — invoke plugin helpers in smoke:**
The synthetic CMakeLists only did `find_package(Pulp REQUIRED)` + a target-existence check. None of the function-body paths where the #2087 piggyback bug lived (`_pulp_add_standalone`'s include of `PulpLinkFontconfig.cmake`) ran. Updated the consumer to actually call `pulp_add_plugin(...)` with a trivial stub source file and CLAP-only formats. This exercises `_pulp_add_standalone` and any downstream-facing helper through the function-body path — exactly where install-time include regressions surface.

**Bonus — launcher resolution fix (caught by manual `/mcp` repro):**
On a fresh `rm -rf ~/.pulp/ && curl install.sh | sh`, restarting Claude Code surfaced `Failed to reconnect to pulp: ENOENT` in `/mcp`. Root cause: Claude Code spawns MCP servers with a sanitized environment, so the launcher's `command -v pulp-mcp` returns empty even when `~/.pulp/bin/pulp-mcp` exists (PATH from the interactive shell doesn't propagate). Added a `${HOME}/.pulp/bin/pulp-mcp` probe between the source-tree build path and the `$PATH` fallback. Verified locally with `env -i HOME=$HOME PATH=/usr/bin:/bin <launcher> --version` → "pulp-mcp 0.102.0" (pre-fix → "cannot locate pulp-mcp binary").

PR #2094, opened 2026-05-15
fix(cmake): ship PulpLinkFontconfig.cmake in SDK + use stable include path (closes #2089)

### LINKED ISSUE #2089: SDK install: lib/cmake/Pulp/PulpUtils.cmake references `${CMAKE_SOURCE_DIR}/tools/cmake/PulpLinkFontconfig.cmake` (downstream-project path, not Pulp's)
## Problem

Downstream `find_package(Pulp ...)` from any plugin (e.g. Spectr) fails at configure because the installed `PulpUtils.cmake` does:

```cmake
# ~/.pulp/sdk/0.101.5/lib/cmake/Pulp/PulpUtils.cmake:997
include(${CMAKE_SOURCE_DIR}/tools/cmake/PulpLinkFontconfig.cmake)
```

`${CMAKE_SOURCE_DIR}` at that point is the **downstream project's** source dir (e.g. `/Users/danielraffel/Code/spectr/`), not Pulp's. So CMake looks for `<spectr-root>/tools/cmake/PulpLinkFontconfig.cmake`, which doesn't exist.

Verified: ` PulpLinkFontconfig.cmake` is **not** in `~/.pulp/sdk/0.101.5/lib/cmake/Pulp/`:
```
$ ls ~/.pulp/sdk/0.101.5/lib/cmake/Pulp/ | grep -i fontconfig
(empty)
```

Adjacent issue I worked around locally: `${CMAKE_CURRENT_LIST_DIR}` doesn't work either if the include sits inside a function body — CMake captures CMAKE_CURRENT_LIST_DIR at function-call time, not function-definition time. The locally-verified working fix is `${Pulp_DIR}/PulpLinkFontconfig.cmake` (Pulp_DIR is set by find_package and points at lib/cmake/Pulp/).

## Reproduction

```bash
pulp sdk install                                         # populates ~/.pulp/sdk/0.101.5/
cmake -S /Users/danielraffel/Code/spectr -B /tmp/spectr \
  -DCMAKE_BUILD_TYPE=Release \
  -DSPECTR_NATIVE_EDITOR=ON \
  -DPulp_DIR=/Users/danielraffel/.pulp/sdk/0.101.5/lib/cmake/Pulp

# Error:
#   include could not find requested file:
#     /Users/danielraffel/Code/spectr/tools/cmake/PulpLinkFontconfig.cmake
#   Call Stack:
#     ~/.pulp/sdk/0.101.5/lib/cmake/Pulp/PulpUtils.cmake:251 (_pulp_add_standalone)
#     CMakeLists.txt:390 (pulp_add_plugin)
```

## Fix (two changes)

1. **Install the helper alongside other cmake files.** Find the existing `install(FILES … DESTINATION lib/cmake/Pulp)` block in root `CMakeLists.txt` and add `tools/cmake/PulpLinkFontconfig.cmake` to the list.

2. **In `tools/cmake/PulpUtils.cmake`, change the include path** from:
   ```cmake
   include(${CMAKE_SOURCE_DIR}/tools/cmake/PulpLinkFontconfig.cmake)
   ```
   to:
   ```cmake
   include("${Pulp_DIR}/PulpLinkFontconfig.cmake")
   ```
   `Pulp_DIR` resolves to `lib/cmake/Pulp/` after install (set by find_package), and to the in-tree directory at build time too. Verified working with a local edit of the installed file.

## Test

Add a CI step that does a "consume from install" smoke: install Pulp to a temp prefix, configure a minimal external project that calls `pulp_add_plugin(...)`, assert the configure succeeds. This is a generally useful pattern beyond this specific bug — catches every "in-tree works, install-mode breaks" regression.

(Could live in `.github/workflows/install-consumer-smoke.yml`.)

## Impact

Blocks any downstream that needs to call `pulp_add_plugin(... STANDALONE ...)` against an installed SDK. Spectr in particular cannot be rebuilt against v0.101.5 SDK without a workaround.

Filed from a real Spectr rebuild attempt 2026-05-15.

danielraffel:
Two-part fix for the broken downstream-find_package consumer flow:

1. Root CMakeLists.txt: add `tools/cmake/PulpLinkFontconfig.cmake` to the
   `install(FILES … DESTINATION lib/cmake/Pulp)` rule so the helper actually
   ships in the SDK tarball (it was being referenced from PulpUtils.cmake but
   never installed).

2. tools/cmake/PulpUtils.cmake: replace the broken
     include(${CMAKE_SOURCE_DIR}/tools/cmake/PulpLinkFontconfig.cmake)
   with
     include("${_pulp_utils_cmake_dir}/PulpLinkFontconfig.cmake")
   where `_pulp_utils_cmake_dir` is captured at file-parse-time (file scope,
   not function scope). The original `${CMAKE_SOURCE_DIR}` resolved to the
   downstream consumer's source tree (e.g. /Users/.../spectr/), not Pulp's,
   so any plugin doing `find_package(Pulp REQUIRED)` and calling
   `pulp_add_plugin(... STANDALONE ...)` hit a configure-time
   "include could not find requested file" error.

   A naïve fix using bare `${CMAKE_CURRENT_LIST_DIR}` inside the function
   body would ALSO break — that variable reflects the caller's current
   list file at function-call time, not where PulpUtils.cmake lives. The
   file-scope capture is the only pattern that survives both contexts.

Repro
-----
v0.101.5 SDK install (the latest published) blocked any downstream from
configuring against an installed SDK on Linux paths. Verified the fix
locally by editing ~/.pulp/sdk/0.101.5/lib/cmake/Pulp/PulpUtils.cmake to
match this PR's pattern — Spectr then configured + built cleanly.

Test
----
test/test_install_consumer_cmake.sh exercises three static invariants
that together pin both halves of the bug:

1. install(FILES …) for lib/cmake/Pulp ships PulpLinkFontconfig.cmake
2. PulpUtils.cmake doesn't include via ${CMAKE_SOURCE_DIR}/...
3. PulpUtils.cmake uses the file-scope `_pulp_utils_cmake_dir` pattern

The release pipeline can layer a heavier "real install + consumer build"
test on top; this is the cheap pre-merge guard that catches this bug
class without a multi-minute build.

4/4 assertions pass locally.

Cross-refs
----------
- #2089 — original report from Spectr-rebuild attempt 2026-05-15
- #1986 — original PulpLinkFontconfig.cmake (same fontconfig fix for pulp-cli)
- #2018 — same fix for pulp-import-design

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -922,6 +922,7 @@ install(FILES
     "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpGenerateWindowsIcon.ps1"
     "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpUtils.cmake"
     "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpPlugin.cmake"
+    "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpLinkFontconfig.cmake"
     "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/FindSkia.cmake"
     "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpInfoPlist.aax.in"
     "${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpInfoPlist.au.in"
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -121,6 +121,24 @@ if(UNIX)
         LABELS "parser-import")
 endif()
 
+# Install-consumer cmake smoke test (pulp #2089). Pins three invariants
+# that together prevent the "in-tree works, install breaks" regression
+# class:
+#   1. install(FILES …) for lib/cmake/Pulp ships PulpLinkFontconfig.cmake
+#   2. PulpUtils.cmake does NOT include via ${CMAKE_SOURCE_DIR}/...
+#      (which resolves to downstream consumer's tree, not Pulp's)
+#   3. PulpUtils.cmake uses the file-scope _pulp_utils_cmake_dir
+#      pattern that survives both in-tree and downstream contexts
+#
+# Cheap static checks — runs in <1s without a real install.
+if(UNIX)
+    add_test(NAME install-consumer-cmake
+        COMMAND bash ${CMAKE_CURRENT_SOURCE_DIR}/test_install_consumer_cmake.sh)
+    set_tests_properties(install-consumer-cmake
+        PROPERTIES TIMEOUT 30
+        LABELS "tooling")
+endif()
+
 # Debug-SDK perf-killer guard smoke (pulp-internal task #35).
 # Hermetic — runs the guard against a tempdir-fabricated SDK install,
 # exercises the three branches (Debug-no-override → FATAL_ERROR,
--- a/test/test_install_consumer_cmake.sh
+++ b/test/test_install_consumer_cmake.sh
@@ -1,0 +1,123 @@
+#!/usr/bin/env bash
+# test_install_consumer_cmake.sh — pin the SDK install layout invariants.
+#
+# Why this test exists
+# --------------------
+# pulp #2089: `lib/cmake/Pulp/PulpUtils.cmake` did
+#   include(${CMAKE_SOURCE_DIR}/tools/cmake/PulpLinkFontconfig.cmake)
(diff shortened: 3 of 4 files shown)


danielraffel:
Superseded by #2091 — that PR includes the same fix as a piggyback alongside the broader #2087 SDK-update UX implementation. Closing this to avoid merge conflicts.

The fix in #2091 is functionally equivalent (file-scope-captured directory pattern; helper added to install rule) plus an additional CI gate (`install-consumer-smoke.yml`) that I had only static-checked. Net better coverage.

PR #2119, opened 2026-05-16
feat(design-import): V2 wire-up — emit registerShortcut + synthetic keydown re-dispatch

danielraffel:
## Summary

V2 wire-up for keyboard shortcut auto-import. Strategy A from codex consult: native chord intercept + synthetic keydown re-dispatch. Pulp owns the OS shortcut capture; the bundled JS owns the React closure semantics.

Builds on #2116 (V1 extractor). Each entry produced by `extract_keyboard_shortcuts` becomes one `registerShortcut(...)` call plus a `__pulpShortcutHandler_N` thunk that calls `__dispatch__('__global__', 'keydown', {...})` so the original React handler in the bundled JS fires with its live closure state.

### Pieces
- **Helpers** in `core/view/`: `key_string_to_keycode` (ASCII passthrough for letters/digits, named-key table for navigation/F-keys/arrows), `modifier_strings_to_mask` (`meta` → `kModCmd` so `metaKey||ctrlKey` cross-platform idiom lands on the right native modifier on macOS and Win/Linux alike).
- **CodeGenOptions** gains `std::vector<DetectedShortcut> shortcuts`. `generate_pulp_js()` emits the `registerShortcut` block after token assignments, before native node construction. Unmapped keys are skipped with a comment for human triage.
- **CLI** in `tools/import-design/pulp_import_design.cpp`: `--no-import-shortcuts` opt-out (default: import); `--shortcuts <path>` output path (default `shortcuts.json`); extraction runs BEFORE parser dispatch while raw TSX is still in scope.
- **Scoping fix** in `extract_keyboard_shortcuts`: `collect_modifiers` walks back to enclosing `if(...)` open-paren instead of a fixed 200-char window. Pre-fix, sibling-branch modifier checks leaked into bare-key matches.

### Tests
16 cases / 69 assertions, all pass locally:
- V1 extractor cases unchanged (9).
- `key_string_to_keycode` + `modifier_strings_to_mask` coverage.
- `generate_pulp_js` emission shape: registerShortcut line, handler thunk body, `__dispatch__` payload.
- Unmapped-key skip, empty-vector no-op, multi-branch scoping pin.

### Manual smoke

```
pulp import-design --from v0 --file synthetic.tsx \
  --output ui.js --shortcuts shortcuts.json
```

Synthetic source exercising bare key / chord / multi-modifier / mode-key / `addEventListener('keydown')` / inline JSX `onKeyDown` produces 5 correctly-mapped shortcuts (Escape, Enter, ArrowLeft no mods; F = alt+shift; s = meta) and a generated `ui.js` containing the matching `registerShortcut` + handler thunks.

## Followups for V3
- End-to-end runtime test: build generated `ui.js` into a `WidgetBridge`, call `forward_key_event` with the chord, assert the React handler fires.
- Spectr roundtrip on the editor fixture.
- Default-shortcuts pass (Settings → `Cmd+,`, Help → `Cmd+?`, etc.) gated on high-confidence component/aria matches.

## Test plan
- [x] `ctest -R design_import_shortcuts` — 16/16 pass
- [x] Local smoke against synthetic TSX — 5 shortcuts, correct chord/keycode mapping
- [ ] CI matrix green
- [ ] Spectr fixture roundtrip (deferred to V3)

### DIFF
--- a/.agents/skills/import-design/SKILL.md
+++ b/.agents/skills/import-design/SKILL.md
@@ -593,3 +593,22 @@ Two safety nets enforce the contract going forward:
 When adding a new live host (e.g. a future `examples/<thing>` binary), update **both**: append the new host to `HOST_FILES` in `host_pump_lint.py` AND to `HOSTS` in `live-host-pump-smoke.sh`. The lint catches source-level regressions at PR time; the smoke catches runtime breakage where the source pairing is correct but the run loop is misconfigured.
 
 For unobtrusive smoke / CI runs, the design-tool exposes `--no-show-window` (uses `WindowOptions.initially_hidden` to skip Dock icon + window display while keeping the full bridge run loop active) and `--exit-after-ms <N>` (clean `request_close()` after N ms). Both flags compose, so `pulp-design-tool --script <probe.js> --no-show-window --exit-after-ms 2000` runs the full live-host code path with no GUI flash.
+
+## Keyboard shortcut V2 wire-up
+
+The import path detects `keydown`/`keyup` global-shortcut handlers in React-style source and emits two pieces of generated code in the host JS bundle:
+
+1. **`registerShortcut(...)` calls** that bind each detected keycode + modifier combo to a synthetic-keydown re-dispatch handler. The native side intercepts the bare key (no DOM focus) and routes through this registration.
+2. **Synthetic keydown re-dispatch** that builds a `KeyboardEvent`-shaped object with the right `key`, `code`, `keyCode`, and modifier flags (`ctrlKey`, `metaKey`, `shiftKey`, `altKey`), then dispatches it to the document so the original React handler's `if (e.ctrlKey && e.key === 's')` branch fires.
+
+Two correctness gotchas the codegen MUST respect — both surfaced by real handler shapes:
+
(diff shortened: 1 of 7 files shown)


danielraffel:
Both Codex review comments addressed in commit `b035ff55b` (was already on the branch); CI gate addressed in commit `0ee49aed1`.

**P1 (line 588) — Ctrl-only shortcuts preserved in modifier mask:**
The collector now emits BOTH `"meta"` and `"ctrl"` when the source uses the cross-platform `e.metaKey || e.ctrlKey` idiom (separately driven by `.metaKey` / `.ctrlKey` substring matches in the surrounding boolean expression, see `core/view/src/design_import.cpp:396-405`). The codegen at L4117-4137 then emits two separate `registerShortcut(...)` bindings — one with `kModCmd`, one with `kModCtrl` — so a Win/Linux-only `e.ctrlKey && e.key === 's'` source handler is registered with the correct mask and the synthetic event sets `ctrlKey: true, metaKey: false` (L4093/4099). Cmd-only and Meta-only source handlers stay single-binding.

**P2 (line 544) — KeyboardEvent.code letter/digit forms decoded:**
The keycode lookup at L534 now strips `KeyA…KeyZ` to the letter and `Digit0…Digit9` to the digit before table indexing. Previously these fell through to keycode 0 and `generate_pulp_js` silently dropped the `registerShortcut(...)` line.

**Tests:** E2E roundtrip in `test/test_design_import.cpp` (commit `6eb001674`) exercises codegen → WidgetBridge → React-style handler, so any future regression of either P1 or P2 fails the suite hard.

**Gate fix in `0ee49aed1`:** added `chore: bump versions` (SDK 0.103.0 → 0.105.0 minor) and a section in `.agents/skills/import-design/SKILL.md` documenting the Ctrl/Meta split + `KeyboardEvent.code` decoding so the next shortcut-shape detector inherits the contract.

PR #2134, opened 2026-05-17
test(view): batch lifecycle widget coverage edges

danielraffel:
## Summary
- add FrameClock coverage for in-tick subscription, self-unsubscribe, and compaction preserving active subscribers
- add WindowManager coverage for message handler replacement, saved visibility state, and shared theme propagation
- add SplashScreen lifecycle coverage for manual dismissal, hidden restart, and repeated show completion
- add Panel paint coverage for theme token colors, fallback colors, and inset border geometry

## Local validation
- `git diff --check`
- `cmake --build build --target pulp-test-frame-clock pulp-test-window-manager pulp-test-splash-screen pulp-test-panel -j$(sysctl -n hw.ncpu)`
- `./build/test/pulp-test-frame-clock "[coverage][phase3]"`
- `./build/test/pulp-test-window-manager "[coverage][phase3]"`
- `./build/test/pulp-test-splash-screen "[coverage][phase3]"`
- `./build/test/pulp-test-panel "[coverage][phase3]"`
- `PULP_DIFF_COVER_CTEST_REGEX='FrameClock subscriber added during tick starts on the next frame|FrameClock subscriber can unsubscribe itself while returning true|FrameClock false-return compaction preserves later active subscribers|WindowManager replacing a message handler drops the previous callback|WindowManager saved state captures visibility and restore preserves other windows|WindowManager shared theme getter and later registration stay in sync|SplashScreen manual dismiss waits for fade-out before callback|SplashScreen show restarts state after hidden dismiss|SplashScreen can be shown again after completion|Panel: custom theme tokens drive fill and stroke colors|Panel: missing theme tokens use documented fallback colors|Panel: border geometry is inset and radius is clamped by border width' tools/scripts/local_diff_cover.sh`

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,4 @@
-# codecov.yml — Codecov config for public/#566 Phase 1 (measurement baseline).
+# codecov.yml - Codecov config for public/#566 Phase 1 (measurement baseline).
 #
 # Primary consumers (in order): Pulp developers, dispatched agents, CI
 # enforcement, contributors reading PR comments. Public dashboard is a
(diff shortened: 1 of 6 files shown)


danielraffel:
Addressed in 48731a7a2: FrameClock::tick now snapshots the active tick window and looks subscribers back up by id after callbacks, so subscribe/unsubscribe during a tick no longer invalidates the active iteration. Re-ran focused view/widget tests and local diff-cover; diff-cover is 83% on the production diff, above the 75% floor.

danielraffel:
Addressed Codex P1 frame-clock comment: this branch already changes `FrameClock::tick` away from range iteration. It snapshots the pre-tick subscriber count, indexes by position only before invoking the copied callback, and re-finds the subscriber by id after the callback returns, so a callback that calls `subscribe()` cannot invalidate an active range iterator/reference. The commented test pins the next-frame subscription behavior against that implementation.

PR #2128, opened 2026-05-17
feat(design-import): Phase A — default shortcuts (source-matched heuristic)

danielraffel:
## Summary

Phase A of the default-shortcuts pass spec'd in `planning/2026-05-16-default-keyboard-shortcuts.md`. On top of #2119 (V2 extractor) and #2120 (focus-guard), this PR adds **source-matched defaults** — auto-bind platform-convention shortcuts when the dev's React source has a high-confidence component match.

Built directly on #2119's existing codegen path (no fork): accepted defaults are lowered into `DetectedShortcut` form and ride the same `registerShortcut(...)` + synthetic-keydown thunk emission.

### What it binds

| Pattern | macOS | Win/Linux | When |
|---|---|---|---|
| Settings / Preferences | `Cmd+,` | `Ctrl+,` | dev's React source has named-modal |
| Help / About / Docs | `Cmd+?` | `F1` | dev's React source has doc-style modal |
| Keyboard cheatsheet | `?` (bare) | `?` (bare) | dev's React source has `<kbd>`-list modal — bare `?` safe via #2120 focus-guard |
| New / Open / Save / Find | `Cmd+N/O/S/F` | `Ctrl+N/O/S/F` | dev's React source has named button |

**What it does NOT bind:** surfaces Pulp doesn't actually provide. Phase B (separate PR) handles Pulp-framework defaults — Cmd+, → Settings tab, Cmd+Opt+A/M → Audio/MIDI sub-tabs in the standalone `SettingsPanel`.

### Hard rule (encoded)

A wrong auto-binding is worse than no binding. Detector requires **≥2 signals** to fire:
1. Component name keyword match
2. `role="dialog"` / `aria-label="..."` in body
3. `<h1>`/`<h2>` heading text in body
4. `<kbd>` tag (cheatsheet disambiguator)
5. **Canonical-name bonus** (added after smoke-test): exact `<Pattern>{Modal,Dialog,Panel,Popover,Sheet,Window,Drawer}` shapes count as 2 signals on their own — real apps like Spectr use inline-styled divs without ARIA, but `SettingsModal` / `HelpPopover` are structurally unambiguous

Ambiguous matches (multiple candidates) → no bind, emit `collision` entry instead. Extracted shortcuts (V2) always win over same-chord defaults.

### Validation against Spectr's real `editor.generated.tsx`

```
pulp-import-design --from claude --file editor.generated.tsx
  → 2 extracted (Enter→blur at line 2458, Escape→onClose at line 2001)
[...]

PR #2137, opened 2026-05-17
fix(view): WidgetBridge static registry — generic keyboard dispatch (closes platform-key-wireup gap, #2128 follow-up)

danielraffel:
## Summary

- New `WidgetBridge::dispatch_global_key(key, mods, is_down)` static fan-out: every live `WidgetBridge` auto-registers in a mutex-protected set in its ctor (dtor unregisters), and platform hosts call `dispatch_global_key` from both `keyDown:` (bare keys) AND `performKeyEquivalent:` (Cmd-modified chords). No per-app `on_global_key` wiring required.
- Before this PR, `WidgetBridge::forward_key_event` was reached only by `examples/design-tool/main.cpp`'s explicit lambda — meaning every other standalone Pulp app (Spectr, future @pulp/react apps) silently dropped `registerShortcut(...)` and `window.addEventListener('keydown', ...)` delivery. Now any standalone app gets it for free.
- macOS `performKeyEquivalent:` override added so Cmd-modified chords (Cmd+,, Cmd+S, Cmd+?) reach the dispatch path instead of being consumed by NSResponder's first-responder chain.

## Validation

Three regression tests, **51 assertions all green**:

| Test | Assertions | What it pins |
|---|---|---|
| `pulp-test-widget-bridge-dispatch-global-key` | 16 | Fan-out to N live bridges; dtor auto-unregister (no dangling pointer); `is_down=false` correctly drops at per-bridge level |
| `pulp-test-mac-perform-key-equivalent` | 4 | Synthesizes a real Cmd+, NSEvent and invokes `performKeyEquivalent:` on a live `PulpView`; asserts on_global_key fires with `kModCmd` |
| `pulp-test-platform-key-wireup` | 31 | E2E: loads imported `ui.js` from #2128's codegen + verifies all 5 Phase A chords (Cmd+,, Ctrl+,, Cmd+?, F1, Escape) reach a `window.addEventListener('keydown', ...)` listener |

**Real-world Spectr roundtrip** — installed Pulp SDK 0.106.0-wireup with these changes, rebuilt `Spectr.app` against it, confirmed bare-key 'S' shortcut in `editor.tsx` now fires through the chain in the live app. (`WidgetBridge::dispatch_global_key` symbol verified present in the linked `Spectr` executable.)

## Phase B follow-up (not in this PR)

Spectr's native app uses `@pulp/react`'s runtime React reconciler — Phase A defaults from `pulp-import-design` (Cmd+,, Cmd+?) are NOT injected into that runtime path. To make Cmd+, → SettingsModal-open work in Spectr we still need runtime injection of `shortcuts.defaults.json` into `@pulp/react`. Tracked as a follow-up; this PR is foundation only.

## Test plan
- [x] `pulp-test-widget-bridge-dispatch-global-key` passes (16/16 assertions, 3 cases)
- [x] `pulp-test-mac-perform-key-equivalent` passes (4/4 assertions, 2 cases)
- [x] `pulp-test-platform-key-wireup` passes (31/31 assertions, 2 cases)
- [x] Spectr.app rebuilt against SDK with these changes — 'S' fires Spectr's mode-switch handler in live app
- [ ] CI green on all three platforms


danielraffel:
Addressed [Codex P1](https://github.com/danielraffel/pulp/pull/2137#discussion_r) in 4043e6339 (#2137) / 41b354b40 (#2139): switched the static registry to `std::recursive_mutex` and now hold the lock through the entire fan-out (no more snapshot-then-unlock UAF window). Pulp's bridge lifecycle is strictly host-driven (never invoked from JS), so the recursion is defensive — pure safety, no exercised re-entry. All 30 assertions across both regression tests still green.

PR #2101, opened 2026-05-16
fix(view): restore native Dawn GPU bridge in JS preludes (cube renders again)

danielraffel:
## Summary

Pulp's native Three.js demo has been rendering an empty black canvas since `ad7175be7` (2026-04-06) "restore main's web-compat-document.js". That commit reverted `core/view/js/web-compat-document.js` to its pre-Phase-14 form to fix an unrelated nested-appendChild regression — but along with it, removed `createNativeAdapter`, `describeNativeAdapter`, the native-aware `navigator.gpu.requestAdapter`, and the bridge-aware `__createMockGPUDevice` 3rd-arg path. Result: every Three.js draw call landed in a no-op mock GPU device. The JS layer ran end-to-end (rAF loop ticked at 220+ frames) but nothing reached Dawn — the demo's solid-black canvas was the visible symptom, masked by Three.js's `new Promise(async (resolve, reject) => {…})` async-executor anti-pattern that swallows throws.

This PR restores the native bridge plumbing AND fixes two latent bugs that fired the moment the bridge was actually engaged.

## Verification

**Cube renders** — green cube on red WebGPU canvas through native Pulp Dawn. Zero browser, zero WebView. Screenshot: [`planning/screenshots/threejs-cube-rendering-2026-05-15.png`](https://github.com/danielraffel/pulp-planning/blob/main/screenshots/threejs-cube-rendering-2026-05-15.png).

**Spectrum mode** also rendering: [`planning/screenshots/threejs-spectrum-rendering-2026-05-15.png`](https://github.com/danielraffel/pulp-planning/blob/main/screenshots/threejs-spectrum-rendering-2026-05-15.png).

**Diagnostic confirms native dispatch**: `globalThis.__phase13BufferedSkips` is empty post-render → all draws hit `__gpuQueueDrawBufferedImpl` → real Dawn device.

## Changes

- **`web-compat-document.js`** — wholesale restore from `8f2ffd6a2` (the last cube-rendering commit). Adds back `createNativeAdapter`, `describeNativeAdapter`, the native-aware `requestAdapter`, the bridge-aware 3rd-arg `__createMockGPUDevice` / `__createGPUAdapter`, `_nativeBridge` propagation through `__createMockGPUTexture` / `__createMockGPUTextureView`, `device.createTexture` routing through `__gpuCreateTextureImpl`, and `pushErrorScope` / `popErrorScope` (required by Three.js's `WebGPUPipelineUtils`).
[...]

PR #2139, opened 2026-05-17
feat(view): Esc closes React popovers via document outside-click fan-out (#2128 follow-up)

danielraffel:
Stacks on #2137 (platform-key-wireup foundation).

## Summary

Pressing Esc in any standalone Pulp app now closes React popovers/dropdowns that use the standard `document.addEventListener('mousedown', onDoc)` click-outside idiom — no per-app wiring, no per-popover `role=` attribute required.

Three pieces, framework-only, **no Spectr code touched**:

1. **`core/view/js/web-compat-document.js`** — `document.addEventListener` / `removeEventListener` / `dispatchEvent` become real (per-event listener map + safe iteration over a snapshot). Previously these were no-ops (pulp #2101, to silence Three.js OrbitControls cleanup); the no-op silently dropped every React click-outside listener in every Pulp standalone. Three.js's cleanup contract is preserved — removeEventListener still tolerates unknown handlers and unknown event types (see test).

2. **`core/view/src/widget_bridge.cpp`** — `__dispatch__` preamble gets a new `id === 'document'` fan-out arm that calls `document.dispatchEvent({type: eventName, ...})`, mirroring the existing `id === '__global__'` window-listener arm. New static `WidgetBridge::dispatch_document_event(type, jsonLiteral)` snapshots the static `all_bridges_` set (same lifecycle as `dispatch_global_key` from #2137) and evaluates the dispatch once per bridge.

3. **`core/view/platform/mac/window_host_mac.mm`** Esc handler — AFTER the existing `dispatch_global_key('Escape', ...)` and BEFORE the modal/ComboBox/`active_overlay_` paths, fire synthetic `pointerdown` + `mousedown` on `document` at coords `(-1, -1)` with `target: null`. The `-1/-1/null` sentinels are outside every real bounding box, so `ref.current.contains(e.target)` correctly resolves to "outside" in every React click-outside handler. Additive — the existing Esc paths still get first crack.

## Test plan

- [x] `pulp-test-widget-bridge-dispatch-document-event` passes — 14 assertions across 3 cases:
  - `document.addEventListener` actually fires; `dispatchEvent` reaches the handler with correct fields
  - Static fan-out reaches every live bridge
  - Ephemeral bridge auto-unregisters on dtor (no dangling pointer)
  - `removeEventListener` tolerates unknown handlers and unknown event types (Three.js cleanup contract)
- [x] Real-world Spectr roundtrip: SDK 0.106.0-wireup reinstalled with these changes, Spectr.app re-linked, `dispatch_document_event` symbol present in the linked Spectr binary
- [ ] User-test: opening Spectr's `PickerDropdown` / `ContextMenu` / any popover and pressing Esc closes it
- [ ] CI green on all three platforms

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.109.0
+    VERSION 0.110.0
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/view/include/pulp/view/widget_bridge.hpp
+++ b/core/view/include/pulp/view/widget_bridge.hpp
@@ -84,6 +84,16 @@ public:
     // platform main thread in some embedding contexts.
     static void dispatch_global_key(int key_code, uint16_t modifiers, bool is_down);
 
+    // Static fan-out: fire a document-level event (`mousedown`,
+    // `pointerdown`, etc.) into every live WidgetBridge's JS document.
+    // Used by platform hosts to fire synthetic outside-click events on
+    // Esc so React popovers that use the
+    // `document.addEventListener('mousedown', onDoc)` click-outside
+    // idiom close automatically. `event_json_literal` is a JS object
+    // literal expression INCLUDING braces, e.g. `{clientX:-1,clientY:-1,target:null}`.
+    static void dispatch_document_event(const std::string& event_type,
+                                        const std::string& event_json_literal);
+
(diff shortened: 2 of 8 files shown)


danielraffel:
Addressed [Codex P1](https://github.com/danielraffel/pulp/pull/2139#discussion_r) in 4043e6339 (#2137) / 41b354b40 (#2139): switched the static registry to `std::recursive_mutex` and now hold the lock through the entire fan-out (no more snapshot-then-unlock UAF window). Pulp's bridge lifecycle is strictly host-driven (never invoked from JS), so the recursion is defensive — pure safety, no exercised re-entry. All 30 assertions across both regression tests still green.

PR #2282, opened 2026-05-18
fix(ui-preview): gate font probe on Skia

danielraffel:


danielraffel:
Closing as duplicate of #2281. #2281 carries the same ui-preview Skia guard and already has the version/skill gate passing; keeping this open would burn duplicate CI on Namespace.

PR #2398, opened 2026-05-19
chore(ci): persist ccache + brew + FetchContent on Namespace macOS runners

danielraffel:
## Summary

Acts on Namespace telemetry feedback (2026-05-19): our macOS PR jobs were running cold on every dispatch because the `actions/cache` flow is gated to skip `runner.environment == 'self-hosted'` (designed for the pre-cutover Daniels-MacBook-Pro that kept caches on local disk). After the 2026-05-18 Namespace cutover, that gate meant every Namespace macOS build paid the full Skia / Dawn FetchContent download + cold ccache cost.

This PR adds `namespacelabs/nscloud-cache-action@v1` to mount Namespace-managed persistent volumes only when the runner name starts with `nsc-runner-`. The Daniels-MacBook-Pro fallback (still active when Namespace is saturated) is unaffected — its on-disk caches continue to persist locally.

## Cached paths

| Path | What it caches | Why |
|---|---|---|
| `~/Library/Caches/ccache` | Compiler object cache | Skia / Dawn / mbedtls cold rebuild is the long pole; ccache warm hit makes those near-free. |
| `~/Library/Caches/Pulp/fetchcontent-src` | CMake FetchContent payloads | Skia binaries, Dawn source, mbedtls source — multi-GB downloads currently re-fetched on every PR. |
| `cache: brew` (named profile) | Homebrew package cache | `brew install ccache` becomes ~free on warm runs. |

## Expected impact

Per Namespace's telemetry guidance: **~10-15 min off cold builds** once warm caches build up over the first few main-branch runs. Tested today: the U-9 PR's macOS leg on Namespace was ~11 min from a cold cache; expect to drop to single-digit minutes on subsequent runs.

## What stays the same

- `Daniels-MacBook-Pro` fallback (still online, idle when Namespace has capacity).
- Linux + Windows lanes (still use the standard `actions/cache` flow).
- The existing `Restore ccache (GitHub-hosted)` step is unchanged — its `if:` already correctly excludes the Namespace path.

## Test plan

- [x] YAML validates (`yaml.safe_load`).
- [ ] First post-merge main run on macOS warms the Namespace cache.
- [ ] Next PR macOS lane reads from the warm cache → measurable build-time drop in the workflow log.

## Provenance

Filed at the request of the Pulp maintainer after Namespace shared concurrency + cache-volume telemetry. Docs: https://namespace.so/docs/reference/github-actions/nscloud-cache-action

### DIFF
--- a/.agents/skills/ci/SKILL.md
+++ b/.agents/skills/ci/SKILL.md
@@ -1785,3 +1785,23 @@ Precedence on `workflow_dispatch`:
 3. Local default (`PULP_LOCAL_MACOS_RUNS_ON_JSON`).
 
 Manual `workflow_dispatch` with an explicit selector input still overrides; the fix only changes behavior for dispatches that arrive without one (which is the `shipyard pr` shape).
+
+## Namespace macOS cache volumes
+
+Namespace-hosted macOS runners (label `namespace-profile-generouscorp-macos`, runner names start with `nsc-runner-`) are **ephemeral** — disk is discarded between jobs. The standard `actions/cache/restore@v4` flow is gated to skip `runner.environment == 'self-hosted'` because the original local self-hosted Mac kept caches on disk between runs. After the 2026-05-18 Namespace cutover, that gate would have meant every macOS PR build paid the full Skia/Dawn FetchContent download + cold ccache cost.
+
+`.github/workflows/build.yml` runs `namespacelabs/nscloud-cache-action@v1` early in the macOS path when `startsWith(runner.name, 'nsc-runner-')`. The action mounts Namespace-managed persistent volumes at:
+
+- `~/Library/Caches/ccache` (compiler cache; warm cache delivers ~order-of-magnitude faster rebuilds for Skia/Dawn/mbedtls)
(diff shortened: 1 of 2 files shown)


danielraffel:
Superseded by [#2399](https://github.com/danielraffel/pulp/pull/2399) ("ci(namespace): cache brew + ccache + FetchContent on Namespace runners") which landed at 2026-05-19 19:11 UTC, and the follow-up [#2402](https://github.com/danielraffel/pulp/pull/2402) ("fix(ci): namespace-cache gate accepts nscloud-* labels").

The merged version's approach is better than this PR:

- Gate is `contains(matrix.runs_on_json, 'namespace')` (plus `nscloud-*` after #2402) — more robust than my `startsWith(runner.name, 'nsc-runner-')` check.
- Covers macOS AND Linux ccache + FetchContent paths.
- Keeps `cache: brew` (which I dropped because of a Homebrew Ruby crash I hit once — turned out to be a transient bottle issue; the other agent's version with the same `cache: brew` is passing on recent PRs).

Closing as no-op. The Namespace cache benefit is already live on main.

PR #2511, opened 2026-05-20
feat(render): add Dawn GPU timestamp queries for per-pass GPU timing

danielraffel:
Phase 6.5 of the inspector roadmap. Phase 6.1 surfaced per-pass CPU
wall-time around each render pass's draw calls; that hides the real
GPU-side execution time where perf bugs usually live. This adds the
true GPU clock via WebGPU timestamp queries.

- PassStats keeps `time_ms` (CPU, unchanged for back-compat) and gains
  `gpu_time_ms` + `gpu_time_valid`. RenderPassManager gains
  `set_pass_gpu_time()` and `gpu_timestamps_available()`.
- New gpu_timestamp.hpp: GpuTimestampResolver is the pure, GPU-free
  tick -> nanosecond -> millisecond conversion with the feature-absent
  and non-monotonic-miss rules, so the resolution math is unit-testable
  without a live device. GpuPassTimer owns the Dawn QuerySet /
  PassTimestampWrites / ResolveQuerySet + map-readback lifecycle in
  gpu_timestamp.cpp, built only when WebGPU is present.
- gpu_surface_dawn.cpp requests the `timestamp-query` feature
  best-effort at device creation, gated on Adapter::HasFeature so an
  adapter without it still yields a working (CPU-timed) surface, and
  exposes GpuSurface::gpu_timestamps_supported().
- Performance.getMetrics now emits cpu_time_ms, gpu_time_ms and
  gpu_time_valid per pass plus a top-level gpu_timestamps_available, so
  the inspector perf view can show CPU vs GPU side by side.

Tests: test_render_gpu_timestamp.cpp — 8 cases / 41 assertions
covering ns conversion, non-nanosecond tick periods, feature-absent
fallback, non-monotonic miss, interleaved-buffer decode, short-buffer
safety, and RenderPassManager GPU-time integration. Header-only so it
runs without a GPU; the live-device GpuPassTimer path is deferred to
CI's GPU matrix.

### DIFF
--- a/core/render/CMakeLists.txt
+++ b/core/render/CMakeLists.txt
@@ -9,6 +9,7 @@ target_include_directories(pulp-render
 
 target_sources(pulp-render PRIVATE
     src/gpu_surface_dawn.cpp
+    src/gpu_timestamp.cpp
     src/gpu_compute.cpp
     src/skia_surface.cpp
     src/render_loop.cpp
--- a/core/render/include/pulp/render/gpu_surface.hpp
+++ b/core/render/include/pulp/render/gpu_surface.hpp
@@ -78,6 +78,15 @@ public:
     /// Returns an unavailable record when no native adapter has been initialized.
     virtual AdapterInfo adapter_info() const = 0;
 
+    /// Whether the initialized device supports WebGPU timestamp queries
+    /// (the `timestamp-query` feature). Phase 6.5 uses this to decide
+    /// whether the inspector perf view can show real GPU-side pass
+    /// durations or must fall back to the Phase 6.1 CPU numbers. The
+    /// feature is requested best-effort at device creation; an adapter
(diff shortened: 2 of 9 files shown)


danielraffel:
Superseded by #2516 ("add Dawn GPU timestamp queries"), which merged to main. This branch was an earlier, interrupted attempt at the same Phase 6.5 work — closing as a duplicate rather than rebasing a dead branch.

danielraffel:
Closing as superseded. #2516 merged to main and already landed Dawn GPU timestamp queries. This PR is a parallel implementation of the same feature (gpu_timestamp singular naming, gpu_surface_dawn.cpp) and now conflicts with main — re-landing a second implementation of an already-shipped feature isn't viable. If it has improvements #2516 lacks, reopen as a follow-up diff against current main instead.

PR #2544, opened 2026-05-21
fix(render): populate resolved GPU timestamps in read_back() (Codex P1)

danielraffel:
Follow-up to #2516 (Codex P1 review finding).

The Dawn GpuTimestamps implementation created the QuerySet and the
resolve/readback buffers in begin_frame(), but nothing ever ran the
ResolveQuerySet + buffer-map path that fills impl_->last_resolved. So
read_back() always returned an empty vector even when timestamp-query
was supported and init succeeded — GPU timings never surfaced.

Wire the missing path:

- Add GpuTimestamps::resolve(void* dawn_instance_handle): encodes
  ResolveQuerySet from the timestamp QuerySet into the resolve buffer,
  copies it into the host-mappable readback buffer, submits, then
  synchronously map-reads the uint64 ticks (pumping ProcessEvents with
  the same 2s deadline gpu_compute.cpp's readback uses).
- Add decode_resolved_ticks(): a pure, Dawn-free helper in the header
  that turns mapped buffer bytes into a tick vector via alignment-safe
  memcpy. resolve() decodes through it; this is the unit-testable seam.
- read_back() now returns the vector resolve() populated. The CPU-only
  stub resolve() is a no-op returning false; degrades gracefully when
  timestamp-query is unsupported or the device/instance is absent.

Tests: extend test_gpu_timestamps.cpp with decode_resolved_ticks cases
(full decode -> resolve_pass_timings round-trip, 64-bit values, null/
short buffers, partial-tail truncation) and a CPU-only resolve(nullptr)
degradation assertion — the exact gap the P1 calls out. Live-device
QuerySet resolution stays deferred to CI's GPU matrix.

### DIFF
--- a/core/render/include/pulp/render/gpu_timestamps.hpp
+++ b/core/render/include/pulp/render/gpu_timestamps.hpp
@@ -2,6 +2,7 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <cstring>
 #include <optional>
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


danielraffel:
Superseded by #2546 ("fix: address post-merge review follow-ups"), which fixes the same Codex P1 on `gpu_timestamps.cpp` (resolve/readback path populating last_resolved) as part of one comprehensive post-merge review-follow-up PR. Closing this single-finding PR as a duplicate to avoid redundant CI.

PR #2547, opened 2026-05-21
fix(view): address Codex P1/P2 findings on lock-to-source + token-lock

danielraffel:
Three review findings on the merged inspector-roadmap PRs:

- P1: escape_js_single() in lock_to_source.cpp only escaped backslash
  and single-quote, so a tweak value carrying newlines/tabs/other C0
  control chars produced syntactically invalid generated JS. Now
  escapes \n/\r/\t by name and any remaining char < 0x20 as \xNN.
- P2: lock_property_to_style_name() rejected backgroundGradient tweak
  paths. Web-compat codegen represents gradient fills as
  el.style.background, so backgroundGradient now aliases to background.
- P2: lock_token_in_designmd() tracked quoting as a bool and
  render_scalar always emitted double quotes, rewriting a
  single-quoted token as double-quoted. Quote tracking now records the
  actual original quote character and re-emits it verbatim.

Each fix ships with Catch2 coverage in test_lock_to_source.cpp /
test_token_lock.cpp. import-design SKILL.md gains the escaping and
property-alias gotchas.

Also adds the tracked `compat/` directory to ALLOWED_ROOT_PATHS in
source_tree_pollution_check.py — it was added by #2531 (compat.json
split into per-surface parts) without the companion allowlist update,
which blocks every push until corrected.

Follow-up to #2470 and #2508 (Codex review findings).

### DIFF
--- a/.agents/skills/import-design/SKILL.md
+++ b/.agents/skills/import-design/SKILL.md
@@ -468,7 +468,12 @@ token export) are separate phases and are NOT in this engine.
   name) onto the camelCase `el.style.<name>` surface. Hyphen/snake
   fragments camelCase. The allow-list is exactly the set of properties
   `generate_node()` emits — an unknown / mistyped path reports
-  `unsupported_property` instead of writing a bogus assignment.
+  `unsupported_property` instead of writing a bogus assignment. A few
+  leaves *alias* onto a different `el.style.<name>` than their camelCase
+  spelling: `backgroundGradient` maps to `background` because web-compat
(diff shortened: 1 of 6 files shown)


danielraffel:
Superseded by #2546 ("fix: address post-merge review follow-ups"), which fixes the same Codex findings on lock-to-source (control-char escaping P1, backgroundGradient mapping P2) and token-lock (quote-style preservation P2) as part of one comprehensive post-merge review-follow-up PR. Closing as a duplicate to avoid redundant CI.

PR #3130, opened 2026-05-29
fix(scan-worker): preserve symlink identity in bundle filtering

danielraffel:
Codex P2 on #3110: `normalized_bundle_path` used
`std::filesystem::weakly_canonical`, which resolves symlinks. When the
scan directory contains multiple `.clap`/`.vst3` symlinks pointing at
the same real bundle, every alias normalizes to the same real path, so
`same_bundle_path` matches the requested path against ANY of the
sibling aliases. `IsolatedPluginScanner::scan` consumes only the first
JSON line emitted by the worker, so the worker can return or blacklist
a different sibling than the one it was asked to scan, depending on
sort order.

Switch to lex-only normalization: `fs::absolute(...).lexically_normal()`
collapses `.`/`..` and resolves to an absolute path WITHOUT following
symlinks, so `/path/foo.clap` and `/path/bar.clap` (both symlinks to
`/elsewhere/real.clap`) compare distinct.

### DIFF
--- a/test/test_audio_tools.cpp
+++ b/test/test_audio_tools.cpp
@@ -264,8 +264,9 @@ TEST_CASE("audio model registry rejects unsafe Hugging Face file paths",
                                    + static_cast<char>(0x1f) + "model.pt").empty());
     REQUIRE(resolve_checkpoint_url("hf://org/repo/\rmodel.pt").empty());
     REQUIRE(resolve_checkpoint_url("hf://org/repo//model.pt").empty());
-    REQUIRE(resolve_checkpoint_url("hf://org/repo/../model.pt")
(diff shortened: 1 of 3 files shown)


danielraffel:
Closing as fully redundant — both fixes in this PR were independently merged via parallel PRs:
- hf:// `..` rejection landed in #3131 (along with Windows-safe model.json assertions)
- scan-worker symlink-identity fix landed in #3132

Diff against current main is empty. The Codex P2 sweep these fixes addressed is complete:
- #3110 → fixed via #3132
- #3118 → fixed via #3131
- #3105 → already had been fixed before; #3131 added more hardening

No action needed.

PR #3162, opened 2026-05-30
fix(view+ios): correct Three.js bundle path in .appex (Resources/threejs → threejs)

danielraffel:
## Summary

iOS-D.3b Slice 3 (PR #3156) writes the build-time Three.js IIFE
bundle to `<appex>/Resources/threejs/three.iife.js`. That subpath is
correct for macOS bundles but wrong for iOS — app extensions are
flat, so `[NSBundle resourcePath]` IS the bundle root. The matching
loader in `threejs_resources_apple.mm` uses
`pathForResource:ofType:inDirectory:@\"threejs\"`, which on iOS
resolves to `<appex>/threejs/three.iife.js`, **not**
`<appex>/Resources/threejs/three.iife.js`.

Result on a real iPad walk-through: NSBundle silently returns nil,
`pulp::view::threejs_iife_source()` returns `std::nullopt`, and JSC
can never load Three.js even though the bundler script ran cleanly at
build time. The macOS-only required CI lane cannot catch this because
the macOS path uses V8 + ESM, not the iOS IIFE-via-NSBundle path.

## Changes

- `tools/cmake/PulpAuv3.cmake` — POST_BUILD writes to
  `$<TARGET_BUNDLE_DIR:${target}_AUv3>/threejs/` instead of
  `.../Resources/threejs/` (three places: iife output, shim output,
  make_directory target). Comment expanded with a pointer back to PR
  #3156 and a one-line explanation of iOS flat-bundle layout.
- `core/view/include/pulp/view/threejs_resources.hpp` — doc comments
  updated (two places) to reflect the iOS flat layout.
- `core/view/src/threejs_resources_apple.mm` — header comment updated.
  Loader code is unchanged (it was already correct — only the
  build-time output path was wrong).
- `test/test_threejs_resources.cpp` — fixture-path comment updated.

## Test plan

- [x] CMake-paths-only change; no .cpp logic changed → no diff-coverage delta.
- [x] `skill_sync_check.py --mode=report` → no mapped paths touched.
- [x] `version_bump_check.py --mode=hint` → no bump needed.
- [ ] Real iOS .appex build verifies the file lands at
      `<appex>/threejs/three.iife.js` and NSBundle locates it.
- [ ] Optional follow-up: regression test that stages a fixture under
      the unit-test bundle's `threejs/` path and asserts
      `threejs_iife_source().has_value()` on iOS Simulator.

## Discovery

Found during deep-review sweep on iOS-D.3b slices 2-5 while Codex was
over GraphQL quota. The cross-platform CI matrix doesn't exercise the
NSBundle lookup path against an actual iOS .appex layout, so the bug
slipped through. macOS bundles handle `Resources/` correctly via
`Contents/Resources/`, but iOS doesn't — it's the kind of bug that
only surfaces on the iOS device walk-through that slice 3 ships
toward.

### DIFF
--- a/core/view/include/pulp/view/threejs_resources.hpp
+++ b/core/view/include/pulp/view/threejs_resources.hpp
@@ -6,8 +6,9 @@
 // embedded inside the AUv3 `.appex` (currently iOS). The IIFE bundle
 // is produced at build time by
 // `tools/scripts/bundle_threejs_for_jsc.mjs` and copied into
-// `<appex>/Resources/threejs/three.iife.js` by the iOS AUv3 CMake
-// rules.
+// `<appex>/threejs/three.iife.js` by the iOS AUv3 CMake rules. (iOS
+// bundles are flat — `<appex>/Resources/` is a macOS-only path that
+// NSBundle does not search on iOS.)
 //
 // On non-Apple-mobile platforms this returns `std::nullopt` — the
 // macOS / desktop lanes resolve Three.js source files via the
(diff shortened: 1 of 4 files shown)


danielraffel:
Consolidated into #3163 (slice 6 program closeout). The fix is cherry-picked there with the version bump, keeping the iOS-D.3b program close-out in a single PR rather than two near-simultaneous version-bump PRs.

PR #3168, opened 2026-05-30
fix(view): gate gpu_surface_ method calls in widget_bridge.cpp behind PULP_WIDGET_BRIDGE_HAS_GPU_SURFACE

danielraffel:
iOS-D.3b Slice 4 (#3157) added two `gpu_surface_->has_surface()` calls at the
top of `__gpuCanvasConfigureImpl` and `__gpuCanvasDescribeCurrentTextureImpl`
without honoring the PULP_WIDGET_BRIDGE_HAS_GPU_SURFACE gate that the rest of
the file uses for `render::GpuSurface` member access. When PULP_ENABLE_GPU=OFF
(iOS Simulator default, sanitizer matrix), `core/render` is never added as a
subdirectory, the gpu_surface.hpp include path is not propagated, the class
stays forward-declared, and the call fails to compile with:

  error: member access into incomplete type 'render::GpuSurface'

This broke iOS Simulator AUv3 builds end-to-end from origin/main. The macOS
required build lane defaults to PULP_ENABLE_GPU=ON, so it never compiled the
no-GPU translation unit; the macOS sanitizer matrix that DOES set GPU=OFF is
advisory-only and the `cmake-ios-auv3-configure` end-to-end test is tagged
`slow` and excluded from PR validation.

Wrap both call sites in `#if PULP_WIDGET_BRIDGE_HAS_GPU_SURFACE` and default
`presentable=false` when the render module is unavailable. The semantics are
correct: a no-GPU configure has no swapchain to address, so presentable=false
is the honest answer. Closes the PR-validation gap that let #3157 ship with
this defect.

Regression test:
- `pulp-test-widget-bridge-no-gpu-gates` performs a fast static scan of
  widget_bridge.cpp, tracking the preprocessor gate state, and fails if any
  `gpu_surface_->` dereference sits outside a PULP_WIDGET_BRIDGE_HAS_GPU_SURFACE
  or PULP_HAS_SKIA gate. Runs in the default macOS PR lane in milliseconds —
  no Xcode, no iOS toolchain required. Verified bidirectionally:
    * Original buggy code → test fails, naming both offending lines.
    * Patched code → test passes (3 assertions).

Compat-Update: skip prefix=canvas2d reason="no surface change, compile gate fix only"
Compat-Update: skip prefix=css reason="no surface change, compile gate fix only"
Compat-Update: skip prefix=html reason="no surface change, compile gate fix only"
Compat-Update: skip prefix=imports reason="no surface change, compile gate fix only"
Compat-Update: skip prefix=react reason="no surface change, compile gate fix only"
Compat-Update: skip prefix=rn reason="no surface change, compile gate fix only"
Compat-Update: skip prefix=yoga reason="no surface change, compile gate fix only"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.24)
 
 project(Pulp
-    VERSION 0.281.0
+    VERSION 0.281.1
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/view/src/widget_bridge.cpp
+++ b/core/view/src/widget_bridge.cpp
@@ -8291,7 +8291,19 @@ void WidgetBridge::register_api() {
         // configure() returning `presentable: true` is the program's
         // contract that JS draws WILL hit the visible swapchain rather
         // than a silent offscreen texture (Codex pass-1 finding #3).
+        //
+        // `has_surface()` is a member of render::GpuSurface; the full type
+        // is only visible when pulp/render/gpu_surface.hpp is on the
+        // include path. On no-GPU configures (e.g. iOS Simulator with
+        // PULP_ENABLE_GPU=OFF) the render module isn't built, the header
+        // isn't reachable, and GpuSurface stays forward-declared — so we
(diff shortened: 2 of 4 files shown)


danielraffel:
Consolidated into #3163 (slice 6 program closeout). The fix + regression test are cherry-picked there alongside the other slice-2-5 sweep findings, keeping the iOS-D.3b program closeout and its sweep follow-ups in a single PR.

Issue #3217, opened 2026-05-30
iOS Sim: Three.js cube renders to offscreen + poisons Dawn device (firstInstance>0)


danielraffel:
## Summary

iOS-D.3c follow-up. After landing every viable mitigation, the Three.js cube is **NOT** visible in the iOS Simulator AUv3 editor — neither inside the canvas card nor anywhere else — even though the WebGPU pipeline initializes end-to-end and Three.js reports successful per-frame submission at 120 FPS.

The Pulp View tree disappears too (pure-black editor), and we now have the precise Dawn validation message that explains why. This issue documents the exact blocker and the fixes that did and did not work, so the next session can pick it up cold.

## Reproduce

1. Build `PulpThreeJsDemo_HostApp_Embed` for iPad Pro 13-inch (M5) Sim (UDID `35721625-2DFD-45B0-8FBE-B3E96D34EF9B`).
2. Install + launch the host with `--auv3-autoload`.
3. Capture screenshot via XcodeBuildMCP → pure black.
4. Capture unified-log `eventMessage CONTAINS 'PULP' OR 'GpuSurface'`.

## Adversarial control test — PROVES Sim AUv3 GPU compositing works

`PulpGpuSmoke_HostApp_Embed` (pure Skia 2D, no JS, no WebGPU bridge) renders perfectly in the SAME Sim AUv3 host — dark blue background + rotating red square clearly visible. So the iOS Sim CAMetalLayer → AUv3 PluginKit XPC compositing path is **NOT** the blocker.

## What proven works (logs)

- `GpuSurface: created Metal surface from CAMetalLayer`
- `GpuSurface: enabling indirect-first-instance (Three.js / non-zero firstInstance draws)`
- `GpuSurface: device features check: IndirectFirstInstance=ENABLED TimestampQuery=MISSING`
- `GpuSurface: Dawn initialized (surface: presentable)`
- `SkiaSurface: Graphite initialized on shared Dawn device`
- `script-ui[info] PULP_THREE_DEMO: probe adapter resolved (adapter=object, features=function)`
- `script-ui[info] PULP_THREE_DEMO: probe device resolved`
- `script-ui[info] PULP_THREE_DEMO: renderer.init ok`
- `PULP_WEBGPU_BRIDGE: canvas.getContext('webgpu') ok (presentable=true, canvas=__el_6__)`
- `PULP_WEBGPU_BRIDGE: context.configure ok (format=bgra8unorm, size=456x500)`
[...]

PR #3228, opened 2026-05-31
fix(view): render Three.js cube on iOS AUv3 GPU path (#3217)

danielraffel:
## Summary

Closes the WebGPU bridge gap so Three.js's `WebGPURenderer` draw calls actually reach Pulp's native Dawn pipeline AND get blitted into Skia's canvas. Proven end-to-end on iOS Simulator — magenta-clear diagnostic produces visible magenta inside the AUv3 editor's canvas card (reverted before commit). Companion to merged PR #3219.

## Six structural fixes (each was a real bug)

1. **`SkiaCanvas::draw_native_dawn_texture()` stub → real impl** (`core/canvas/src/skia_canvas_shaders.cpp`). Uses `skgpu::graphite::BackendTextures::MakeDawn(WGPUTexture)` + `SkImages::WrapTexture` + `canvas->drawImageRect` per Codex's recommendation in issue #3217.

2. **`texture.createView()` mock didn't propagate `_nativeBridge`** (`core/view/js/web-compat-document-gpu-mock.js`). The buffered shim's `createBufferedDrawPayload` check for `attachmentView._nativeBridge` therefore always failed → all Three.js draws skipped.

3. **`context.getCurrentTexture()` didn't stamp `_nativeCanvasId`** on the bridged texture (`core/view/js/web-compat-canvas-gpu.js`). Without it the new view forwarding from (2) had no canvas id to attach.

4. **`beginRenderPass()` had no `onEnd` callback** (`core/view/js/web-compat-document-gpu-mock.js`). Render-pass commands recorded by the shim were dropped. Added `renderCommands.push` callback + attached the array to the command buffer in `finish()`.

5. **Buffered `encoder.draw` fell back to a no-op when `init.onEnd` was missing** (`core/view/js/web-compat-gpu-buffered.js`). Added a sync-dispatch fallback that calls `__gpuQueueDrawBufferedImpl` directly, both for `draw` and `drawIndexed`.

6. **Dawn validates Skia/Graphite's per-frame instanced draws as `First instance must be zero` on iOS Sim's Metal SoftwareRenderer** even with `IndirectFirstInstance` enabled (`core/render/src/gpu_surface_dawn.cpp`). Enable `skip_validation` + `allow_unsafe_apis` Dawn toggles when `TARGET_OS_SIMULATOR` is true; production iPad keeps validation on. Also filter the known-benign noise from `SetUncapturedErrorCallback`.

## Protective indirect-draw conversion

`core/view/src/widget_bridge.cpp` — at three `pass.Draw` sites, when `first_instance > 0 && device.HasFeature(IndirectFirstInstance)`, convert direct `Draw` → `DrawIndirect` with an inline args buffer. No behavior change today; prevents regression if a real `firstInstance > 0` direct draw ever reaches the bridge.

## Verification

- iOS Sim (iPad Pro 13" M5): View tree + canvas card visible at 120 FPS. Magenta-clear diagnostic produces visible magenta in the canvas card → full pipeline (Three.js → JS hook → native bridge → wgpu::Texture → Skia blit → AUv3 composite) works end-to-end.
- Sim's actual Three.js cube still invisible at the final Metal layer (Sim Metal SoftwareRenderer silently drops Three.js's actual draws). Real iPad device test pending.

## Remaining gap (separate issue)

iPad first-install AUv3 discovery still fails after PR #3219's `instantiationInFlight` + notification observer + 60-poll Task. Will be investigated using Apple's `IncorporatingAudioEffectsAndInstruments` AUv3Host as a known-good comparator host.

## Test plan

- [x] Builds for iOS Sim (Release, iPad Pro 13" M5 destination)
- [x] Builds for iOS device (Release, generic/platform=iOS)
- [x] iOS Sim AUv3 renders View tree, WebGPU pipeline initializes, native bridge dispatches at expected rate
- [x] Sim magenta-clear diagnostic proves end-to-end blit
- [ ] iPad device cube rendering — pending AUv3 discovery investigation


danielraffel:
Marking draft. This branch carries the iOS-D.3c GPU-present wiring + canvas-alpha + (now-duplicated) writeBuffer fix + an os_log subsystem commit. **It is not mergeable yet**: the editor canvas renders **black** because the live iOS present path falls back to an offscreen Skia surface (root cause pinned in #3217). 

Next steps before this can ship:
1. Rebase onto current main (drops the writeBuffer commit, now landed via #3273).
2. Diagnostic iOS build with the new os_log subsystem to name which `SkiaSurface::begin_frame()` fallback branch fires.
3. Fix the present path so the cube reaches the CAMetalLayer drawable; add a headless test that drives the live CanvasWidget::paint path.
4. Visual proof (rendered-cube screenshot) before un-drafting.

PR #3318, opened 2026-06-02
feat(ship): one-off sign+notarize via `share`, `--path` primitives, release notarizes its artifact

danielraffel:
## What

Adds the composable + opinionated paths for "I built something and want to hand it to a friend, signed and notarized" without the full release pipeline.

- **`pulp ship sign --path <app|dmg|bundle>`** — sign one explicit artifact instead of scanning build dirs. Rejects `.pkg` with a productsign pointer.
- **`pulp ship notarize --path <dmg|pkg|app>`** — notarize + staple explicit artifacts (repeatable); falls back to the bundle scan when absent.
- **`pulp ship share <app|dmg|pkg>`** — one-shot: sign → wrap `.app` in a DMG → sign DMG → notarize → staple → `spctl` Gatekeeper verify. `--dry-run` prints the plan. This closes the "Unnotarized Developer ID" Gatekeeper rejection (valid Developer ID signature, no notarization ticket).
- **`pulp ship release --dmg/--pkg`** now notarizes + staples the distributable it packaged (selected by build time), not the loose bundles — so the artifact in `artifacts/` is Gatekeeper-ready and the command name is honest.

## Safety / compatibility

`sign`/`notarize` are backward compatible (identical behavior without `--path`). The production GitHub release pipeline (`.github/workflows/sign-and-release.yml`) does its own `codesign` + `pkgbuild`/`productbuild` + `xcrun notarytool submit` + `stapler staple` and **never invokes these CLI subcommands**, so regular releases are unaffected.

## Tests

9 new shellout cases in `test_cli_ship_shellout.cpp` covering `sign --path` (missing / `.pkg`), `notarize --path` dry-run argv, `release --dmg` actually notarizing the built DMG (real `hdiutil` path), and `share` usage / not-found / bad-ext / dry-run / no-identity + the non-macOS guard. Full ship suite green locally (408 assertions, 45 cases).

## Docs / skills

`cli-commands.yaml`, `docs/reference/cli.md`, `ship` SKILL.md, `cli-maintenance` SKILL.md, and the `/ship` slash command updated with the share flow, `--path` primitives, and the one-off-vs-release distinction.

### DIFF
--- a/.agents/skills/cli-maintenance/SKILL.md
+++ b/.agents/skills/cli-maintenance/SKILL.md
@@ -275,6 +275,32 @@ Reference implementation: PR landing `feature/asc-notary-key-flow`
 `tools/cli/cmd_ship.cpp` (`notarize` block), `test/test_notary_env.cpp`,
 `test/test_cli_ship_shellout.cpp` ASC-key cases.
 
+### A subcommand that orchestrates other subcommands (`pulp ship share`)
+
+`cmd_ship` calls itself recursively (`cmd_ship({"notarize", "--path", ...})`)
+to compose stages — `share` and `release` both reuse the `notarize` handler's
+credential-resolution chain rather than duplicating it. When you add a flag
+like `--path` to a leaf subcommand, any orchestrating subcommand can thread it
+through for free. Conventions that kept this testable without Apple creds:
+
+- Add a `--dry-run` to orchestrators (`share`) that prints the plan and returns
+  0 before any `codesign`/`notarytool`/`spctl` call — shellout tests assert the
+  plan text. Mirror the existing `notarize --dry-run` / `auv3-xcodeproj
+  --dry-run` pattern.
+- Put guard rails (input not found, unsupported extension, missing identity)
+  *before* the side-effecting work so they're exercised by credential-free
+  shellout tests.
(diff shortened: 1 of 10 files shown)


danielraffel:
Addressed both Codex findings in e7918ff65:

- **P1 (sign .pkg before notarizing).** Added `--installer-identity` (Developer ID Installer) to `package` + `release` so the flat package is signed via `pkgbuild --sign`. `release` now also signs standalone `.app`s and the produced `.dmg`. As a safety net, the notarize routing verifies each artifact's signature (`pkgutil --check-signature` / `codesign --verify`) and **skips — never submits — any unsigned `.pkg`/`.dmg`**, with a clear "skipping unsigned artifact" message, instead of falling back to notarizing loose bundles.
- **P2 (reject/​wrap raw .app).** `notarize --path Foo.app` is now rejected up front (notarytool only accepts `.dmg`/`.pkg`/`.zip`) with a pointer to `pulp ship share` (wraps the app in a DMG) or a `ditto` zip recipe.

Tests: added the raw-`.app` rejection + `--installer-identity` parser cases, and rewrote the `release --dmg` test to assert the unsigned-artifact guard. Full ship suite green (427 assertions, 46 cases). Help text, `cli-commands.yaml`, and the ship skill updated.

PR #6115, opened 2026-07-14
feat(signal): joint-channel block oversampling + elliptic half-band IIR

matthewfudge:
Two additions to `pulp::signal`'s oversampling, both from hitting real limits in production DSP. Purely additive — no existing `Kind`, default, coefficient table or API touched, and the new `Kind` is appended so existing ordinals are unchanged.

## Gap 1 — `OversamplerT` can't express joint-channel processing

`OversamplerT` fuses upsample → callback ×N → decimate into one call per input sample, with no way to view the intermediate oversampled buffer. That makes a whole class of standard DSP impossible:

- a **dither draw shared across L/R** at the same oversampled sub-index
- a **stereo-linked compressor sidechain** (`0.5*(|L|+|R|)` driving one shared envelope)

Two independent instances can't do it: each finishes its whole N-sample loop before returning, so channel R's oversampled sample at index *i* doesn't exist while channel L's callback for *i* is running. Any oversampled effect with cross-channel coupling — linked dynamics, mid/side work, correlated dither — needs the split.

**Added:** `polyphase_block_oversampler.hpp` — `PolyphaseBlockOversamplerT<SampleType, UpStage, DownStage>`, a **sibling** class (not new methods on `OversamplerT`) with `process_up()`/`process_down()` and an `OversampledBlockView` callers read/write jointly before decimating. Filter family via template-template params, defaulting to the existing half-band stages.

`OversamplerT` is untouched deliberately: its fused, allocation-free per-sample callback is the right API for the common single-channel case, and folding both shapes into one class would burden every existing caller with surface most don't need. Happy to reshape if you'd rather it lived on `OversamplerT`.

The test `"joint-channel work between up and down is possible"` in `test_polyphase_block_oversampler.cpp` is the argument for the whole file — it does something the fused API genuinely cannot.

## Gap 2 — only one minimum-phase half-band design is available

`halfband_iir.hpp`'s fixed six-section design documents *"group delay ~6 samples at the half-band's input rate"*. That's a good default, but it's the only IIR lane on offer, and its latency/response is baked in. There's no way to trade section count and group delay against transition steepness.

**Added:** `elliptic_halfband_iir.hpp` — `EllipticHalfBandUpsampler2xT`/`DownsamplerT`, same calling convention as the existing half-band classes, running the **Valenzuela–Constantinides elliptic half-band design equations** at `configure()` time (variable section count from the requested stopband/transition spec) rather than a fixed coefficient table. Wired in as `OversamplerT::Kind::elliptic_polyphase_iir`, with `Quality::pristine` selecting the deeper prototype floor.

Both designs are legitimate and they are **not interchangeable** — they differ measurably in group delay, so a caller with a latency budget (or an existing latency contract to hold) currently has no lane to pick. This gives them one.

## Provenance

Both headers implement published filter-design methods (Valenzuela–Constantinides elliptic half-band polyphase-allpass) from Wave Alchemy's own clean-room implementation. Ours to contribute, MIT.

## Tests

`test_elliptic_halfband_iir.cpp` (14 cases), `test_polyphase_block_oversampler.cpp` (5, incl. the joint-channel one), and 7 new/extended cases in `test_oversampling_quality.cpp` (new Kind added to the existing multi-kind loops, plus passband/stopband/reset/group-delay-difference). Registered in `test/cmake/late_core_tests.cmake`. Deterministic inputs only. The new Kind and both Elliptic classes are wired into `test_signal_rt_safety.cpp`'s allocation probe.

Fail-before/pass-after checked rather than assumed: injecting a one-line bug in the downsampler's path-combine (`(holdover_ + direct_out) * 0.5` → `direct_out`) turns both suites red on stopband rejection (-0.0016 dB vs required <-40 dB); reverting turns them green. `ctest` 48/48 for the oversampling-related suites.

Version bumped 0.668.0 → 0.669.0 in its own `chore: bump versions` commit. Two rows added to `docs/reference/modules.md`.

## Note

Independent of #6114 (`FftT::inverse_real()`) — separate branch off `main`, no shared commits. Merge either order.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ include(${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpMacOsArch.cmake)  # macOS ar
 include(${CMAKE_CURRENT_SOURCE_DIR}/tools/cmake/PulpMinOs.cmake)
 
 project(Pulp
-    VERSION 0.668.0
+    VERSION 0.669.0
     DESCRIPTION "Cross-platform audio plugin and application framework"
     HOMEPAGE_URL "https://github.com/danielraffel/pulp"
     LANGUAGES CXX C
--- a/core/signal/include/pulp/signal/elliptic_halfband_iir.hpp
+++ b/core/signal/include/pulp/signal/elliptic_halfband_iir.hpp
@@ -1,0 +1,310 @@
+#pragma once
+
+/// @file elliptic_halfband_iir.hpp
+/// Elliptic (Valenzuela-Constantinides) polyphase-allpass half-band filter for
+/// 2x oversampling — a variable-order, runtime-designed sibling to
+/// `halfband_iir.hpp`'s fixed six-section Vaidyanathan / Regalia-Mitra design.
+///
+/// Both filters realise the same classical "two parallel allpass paths"
+/// half-band decomposition; they differ in *which* elliptic design fills in
+/// the allpass coefficients and in how many sections that design needs for a
+/// requested transition width / stopband floor. `halfband_iir.hpp` pins a
+/// single published six-section design. This header instead runs the
+/// Valenzuela-Constantinides elliptic design equations at configure() time,
+/// so the section count (and therefore the group delay and stopband) tracks
+/// whatever transition width and stopband floor the caller asks for, rather
+/// than being pinned to one fixed operating point.
+///
+/// Group delay is materially different between the two designs at matched
+/// operating points: this elliptic design's per-stage schedule (see
+/// `OversamplerT::Kind::elliptic_polyphase_iir` in oversampling.hpp) lands
+/// close to 0 samples of *extra* passthrough error at the same factor,
+/// where the six-section default in `halfband_iir.hpp` differs by more
+/// than a sample — neither is wrong, they are simply different points on
+/// the latency/transition-width/section-count trade-off, and callers that
+/// need this design's specific latency/response pick this one.
+///
+/// ## Design equations
+/// `design_elliptic_halfband_allpass()` solves for the number of allpass
+/// sections `N` and their coefficients `a_i` from a normalised transition
+/// width (fraction of the input Nyquist) and a stopband floor in dB, via the
+/// elliptic-filter degree equation and the Vaidyanathan/Constantinides
+/// closed form for equiripple half-band allpass coefficients (Valenzuela &
+/// Constantinides, "Digital allpass filter design", and the classical
+/// elliptic-rational-function machinery it builds on). All arithmetic is
+/// double precision; coefficients are narrowed to `SampleType` only at the
+/// end, matching how such designs are computed in every reference
+/// implementation of this method.
+///
+/// ## Section layout
+/// The `N` coefficients solved above are split even/odd-indexed into a
+/// "direct" path (indices 0, 2, 4, ...) and a "delayed" path (1, 3, 5, ...),
+/// each run as a per-sample cascade of first-order allpass sections. This is
+/// the same even/odd split `halfband_iir.hpp` calls path A / path B, just
+/// with a variable, design-dependent number of sections in each path instead
+/// of a fixed six.
+///
+/// RT contract: `configure()` (and the constructor, which calls it) resize
+/// per-section state and allocate. `process()`, `process_block()`, and
+/// `reset()` allocate no memory once configured.
+
+#include <algorithm>
+#include <cmath>
+#include <cstddef>
+#include <span>
+#include <vector>
+
+namespace pulp::signal {
+
+namespace detail {
+
+/// Solve the Valenzuela-Constantinides elliptic half-band allpass design for
+/// a normalised transition width (fraction of the input Nyquist, i.e. `tw`
+/// in `(0, 1)`) and a stopband floor in dB (negative). Returns the flat
+/// coefficient array `[direct sections..., delayed sections...]` that
+/// `EllipticHalfBandUpsampler2xT` / `EllipticHalfBandDownsampler2xT` expect,
+/// i.e. even-indexed `a_i` first, then odd-indexed.
+template <typename SampleType>
+std::vector<SampleType> design_elliptic_halfband_allpass(double normalised_transition_width,
+                                                         double stopband_db) {
+    constexpr double pi = 3.14159265358979323846;
+    const double wt = 2.0 * pi * normalised_transition_width;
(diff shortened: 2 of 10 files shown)


matthewfudge:
Thanks — that was a genuinely useful review. 8 of 9 fixed, 1 rejected with reasoning.

**The enum one was the important catch, and my PR description was wrong.** I claimed "purely additive" — it wasn't. Inserting `elliptic_polyphase_iir` before `linear_phase_fir` shifted that ordinal 2→3, so any persisted integer setting of `2` would silently select a different filter after upgrading. Moved to the end; ordinals are back to 0/1/2 with the new kind at 3. Checked the rest of the diff for other enum insertions — this was the only one. Proof it bites: reverting the fix fails `REQUIRE(linear_phase_fir == 2)` with `3 == 2`.

**Finding 8 was worse than reported.** `stopband_db = -300.0` is a value the doc comment itself advertises, and `stopband_db > -300.0 ? pow(...) : 0.0` forced the gain to exactly 0 at that boundary → `log(0)` → `+inf` order → not just UB on the int conversion: reverting the fix makes the test **hang until SIGKILL**. Removed the special case (`pow(10, -15)` is perfectly fine) and clamped the section count for genuinely pathological inputs.

**Finding 2 — rejected, it's a comment bug not a code bug.** I checked the reference implementation of the design equations: `gain_start + gain_step * n` with a negative start and positive step *shallows* per stage, exactly as the code does. The code is right and the comment was lying. Fixed the comment instead ("deepens" → "shallows off a deep stage-0 floor"). Changing the code here would have introduced a real bug.

**Finding 4 — you were right and I was initially skeptical.** I assumed pulp trusted caller-supplied counts everywhere and that validating only in my new class would be inconsistent with your conventions. Checked: `multi_channel_meter.hpp` clamps `num_channels`/`num_samples` in `process()`, so trusting the caller was the inconsistent choice. `process_up`/`process_down` now clamp to what `prepare()` configured.

Rest: `prepare()` bounds `max_block_size` to `SIZE_MAX >> (kMaxStages+1)` so every subsequent `*2` provably can't overflow (3); `num_stages` clamped to `[0, kMaxStages=24]`, which also keeps `1 << stages` well inside `int` (5, 6); the zero-stage path now uses an owned passthrough scratch instead of `const_cast`-ing your input, and `process_down()` copies it back so out-of-place callers round-trip too (7); `reset()` `std::fill`s instead of assigning, and I wired the new Kind and both Elliptic classes into `test_signal_rt_safety.cpp`'s existing allocation probe — they weren't covered before (9).

`ctest` 48/48 on the oversampling suites; `pulp-test-signal-rt-safety` 19/19.

*(Edited: removed references to another framework by name, per your house rule — applies to my comments as much as the source.)*

matthewfudge:
This one was already compliant — `0282e47d5` is a standalone `chore: bump versions` touching only `CMakeLists.txt`, and `Enforce version & skill sync` is green on it. Looks like the heads-up went to both PRs; no change needed here.

(#6114 genuinely did need it — I'd collapsed its split by amending a fix in. Fixed there as `736f70e62`.)

Noted on the `0.669.0` collision — we'll re-bump whichever of ours lands second.

matthewfudge:
Done — `cfe831208`. All references to another framework are gone from the source, and the sweep turned up more than the seven you spotted: **15 in total**, including two in `test_elliptic_halfband_iir.cpp` and two paragraphs of the commit message itself. `rg -in` across the whole diff and both commit messages is now clean.

Assertion counts unchanged (9431 / 3549 / 5564 / 48) — comments only, as intended.

**I reframed rather than deleted**, since as you say the information a reader needs was already there. The main move was naming the *published method* instead of the framework that also implements it — the design is Valenzuela–Constantinides' elliptic half-band polyphase-allpass method, which is prior art, so:

- `///< ... elliptic design (juce-parity)` → `///< ... elliptic (Valenzuela-Constantinides) design`
- "matching X's specific latency/response" → "the design equations solve directly for a requested transition width and stopband floor, so pick this kind when a specific latency/response operating point matters"
- "`Quality::pristine` maps to X's `isMaximumQuality`" → what pristine actually *does*: narrows the transition width and deepens the stage-0 floor with a larger step
- `docs/reference/modules.md`: "(fixed and elliptic/juce-parity)" → "(fixed six-section and elliptic Valenzuela-Constantinides)"

**One I deleted rather than reframed:** the state-form attribution on the allpass section. Its only content was "this matches the other implementation" — the surrounding sentence already stated the actual fact (transposed direct form, vs. `HalfBandAllpassSectionT`'s direct form I), so nothing was lost. Where a comment was *only* a comparison, it went; I didn't invent a technical claim to fill the space.

**Your house rule is right and I've adopted it as ours** — not just for source, but for PR bodies and review comments too. I've scrubbed both of our open PRs' descriptions and my earlier replies on the same basis. The reason it matters is sharper than style: the framework in question is GPL/commercial dual-licensed and pulp is MIT, so attribution language in a public PR to your repo invites a licensing question about *your* codebase that you should never have to field — even when the implementation genuinely is clean-room. Motivating from the capability gap and the measurement is both safer and, frankly, a better argument.

Ready for another look when you have a moment.

matthewfudge:
Those references are already gone — the review landed against `0282e47d`, and the rewrite went up in `8f193aad1` about nine minutes later. Current head is clean.

I'd rather have caught it than been asked, so I checked the whole surface rather than just the quoted file: source, commit messages, and the PR description all scan clean on the current head, and the same for #6114.

On the substance — agreed, and the guidance was the useful part. Each one now says what the lane *is* and when you'd pick it, rather than what it matches:

```cpp
elliptic_polyphase_iir, ///< Half-band polyphase IIR, elliptic (Valenzuela-Constantinides) design.
```

The header block describes it as a design-equation-driven section count — *"pick this kind when a specific latency/response operating point matters more than matching `polyphase_iir`'s fixed coefficient table."* You're right that the information a reader needs was already there; only the attribution was doing no work.

One judgement call worth surfacing: the design still carries its literature name (Valenzuela-Constantinides). I read that as in-scope, since it names a published design rather than anyone's implementation — but it's your gate and your call, so say the word and it goes too.

The cubic P3 is fixed in the same push, and it was a real doc bug rather than a nit: the schedule comment claimed each stage's stopband floor "deepens", while the code adds a *positive* step to a *negative* floor (`floor_start + step * stage_index`), so later stages get shallower — pristine runs −90/−80/−70. It now reads "shallows by a fixed step off a deep stage-0 floor". Worth having.

Version collision noted — nothing pre-emptive, happy to re-bump whenever this one loses the race.

Ready for another look.

PR #8370, opened 2026-09-14
chore: bump versions

danielraffel:
chore: bump versions

Assigned from Version-Bump intent on c3635e9078cf..3fab348c738f.
sdk 0.854.1->0.854.2 (patch); plugin 0.455.7->0.456.0 (minor)

Version-Bump-Applied: 3fab348c738f61015a78e1f313c5f451573c189c


danielraffel:
## Held as draft — stale range, do not re-arm

This bump was cut against `3fab348c738f` (`Assigned from Version-Bump intent on c3635e9078cf..3fab348c738f`). At the time it was cut that **was** main's tip, so its range was valid. It no longer is: #8363 and #8351 have since merged, and #8312/#8349/#8344 are in the merge queue ahead of it.

Merging it now would strand those merges' version intent. `_first_parent_boundary()` maps this PR's marker commit to its **integration merge**, and the next drain starts *after* that boundary — so every merge that landed between `3fab348c738f` and this PR's merge commit falls outside both ranges and is silently dropped. The queued PRs carry real intent: #8363 and #8312 touch `sdk` trigger paths (`tools/import-design/**/*.cpp`, `experimental/pulp-rs/src/**/*.rs`); #8351 and #8312 touch `plugin` (`.agents/skills/**`).

That is the failure mode documented in `_bump_pr_covers()` from 2026-09-03, and the one that cost a release cycle earlier today.

Draft is the contract-sanctioned hold: `_arm_auto_merge()` treats a draft bump PR as "an explicit release hold" and refuses to arm it. **Please do not mark this ready or re-arm auto-merge.**

Resolution: once the queue drains, **close this PR** and let one `version-at-land` `workflow_dispatch` generate a replacement covering the full `c3635e9078cf..<new main>` range. A held draft also blocks replacement generation, so it should not be left held indefinitely.

